PR #2, opened 2023-07-24
Add initial projects and their structure

stjepangolemac:
This PR is larger than I'd want it to be but I needed quite a large set of types to complete the dependencies for the deposit wireformat and parsing logic.

Initial projects added to the workspace:

- stacks-core
- sbtc-core

It's not using blockstack_lib or stacks_rs because of licensing and completeness issues.

Here's what is added:

- c32 primitives
- crypto primitives
- basic types as principal data and hashes
- StacksAddress
- deposit wireformat and parsing logic
- tests

~~NOTE: Some tests are failing but I'd prefer it if we could fix them in separate PRs. The repo is still new and bare so there's a chance to move a bit faster and unlock tasks that can be done in parallel.~~

I've fixed all tests.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,64 @@
...
+1. Is testing required for this change?
+2. If it’s a bug fix, list steps to reproduce the bug
+3. Briefly mention affected code paths
+4. List other affected projects if possible
+5. Things to watch out for when testing
+
+## Checklist
+- [ ] Code is commented where needed
+- [ ] Unit test coverage for new or modified code paths
+- [ ] `npm run test` passes
```
netrome, line 62:
```suggestion
- [ ] `cargo test` passes
```

sbtc-core/src/wireformat/deposit.rs
netrome, line 1: Is this module only the wire format or is it intended to parse information from a bitcoin transaction as well? Or is that going to be performed in another module? In that case I'd rename `ParsedDepost` to `ParsedDepositData`.

stacks-core/src/address.rs
```diff
@@ -0,0 +1,253 @@
...
+pub enum AddressHashMode {
+    SerializeP2PKH,
+    SerializeP2SH,
+    SerializeP2WPKH,
+    SerializeP2WSH,
+}
+
+#[derive(Debug, Clone)]
+pub struct StacksAddress {
+    version: u8,
```
netrome, line 22: Can we have a strong type for the version instead of a u8?

stacks-core/src/address.rs
```diff
@@ -0,0 +1,253 @@
...
+
+    pub fn hash(&self) -> &Hash160 {
+        &self.hash
+    }
+
+    pub fn from_public_keys(
+        version: u8,
+        public_keys: &[PublicKey],
+        signatures: usize,
+        hash_mode: AddressHashMode,
```
netrome, line 43: I have always found these args slightly confusing. It would be great if we could find a more intuitive interface here. I'm not sure what that would be though, but especially the `signatures` arg is confusing to me. Could we at least have it documented?

stacks-core/src/address.rs
```diff
@@ -0,0 +1,253 @@
...
+impl TryFrom<&str> for StacksAddress {
+    type Error = StacksError;
+
+    fn try_from(address: &str) -> Result<Self, Self::Error> {
+        let (version, hash) =
+            decode_address(address).map_err::<StacksError, _>(|err| err.into())?;
+
+        Ok(Self::new(version, Hash160::new(hash)))
+    }
+}
```
netrome, line 98: Are these traits necessary? Otherwise I'd prefer to rely on the Display and FromStr traits which would be more idiomatic for string representation and parsing.

stacks-core/src/c32.rs
```diff
@@ -0,0 +1,393 @@
+use crate::crypto::hash::SHA256Hash;
+
+const C32_ALPHABET: &[u8; 32] = b"0123456789ABCDEFGHJKMNPQRSTVWXYZ";
+const C32_BYTE_MAP: [Option<u8>; 128] = [
```
netrome, line 4: Could we represent this in a more compact/readable format? Or just create this from the alphabet?

stacks-core/src/crypto/hash.rs
```diff
@@ -0,0 +1,77 @@
...
+impl SHA256Hash {
+    pub fn new(value: impl AsRef<[u8]>) -> Self {
+        let bytes = {
+            let mut buff = [0u8; SHA256_LENGTH];
+
+            let mut ctx = Context::new(&SHA256);
+            ctx.update(value.as_ref());
+
+            let digest = ctx.finish();
+            buff.copy_from_slice(digest.as_ref());
```
netrome, line 20: Would it be possible to avoid copying here?

stacks-core/src/lib.rs
```diff
@@ -0,0 +1,26 @@
...
+
+pub mod prelude {
+    pub use super::StacksError;
+    pub use super::StacksResult;
+
+    pub use crate::address::*;
+    pub use crate::c32::*;
+    pub use crate::crypto::*;
+    pub use crate::utils::*;
+}
```
netrome, line 26: Nit: Personally I'd like to avoid wildcard imports and big preludes. I generally prefer explicit imports.

netrome (approved):
Some misc thoughts and comments, but generally this looks good to me.

.github/PULL_REQUEST_TEMPLATE.md
stjepangolemac: I haven't even looked at that file. I'll reduce it a lot and fix this.

stacks-core/src/address.rs
stjepangolemac: Yes, that's been especially painful now when working on StacksAddress tests.

sbtc-core/src/wireformat/deposit.rs
stjepangolemac: I thought about just putting wireformats and parsing logic in here. I agree with that, will do.

stacks-core/src/address.rs
stjepangolemac: Ah I didn't even see this comment when I wrote https://github.com/stacks-network/sbtc/issues/7

I have a lot of ideas on how to improve it.

stacks-core/src/address.rs
stjepangolemac: Both conversions are fallible, and that's hidden in the `blockstack-lib` implementation with unwraps. I believe this to be closer to the truth. It also prevents the user to be confused, compared to panicking.

stacks-core/src/c32.rs
stjepangolemac: I've changed it so it's much more compact now, and `cargo fmt` won't mess with it. It can be generated but generation is not trivial. I believe there was a comment about generation somewhere. Also, I'm not sure if there are any performance implications when comparing const data and something lazily generated.

```rust
#[rustfmt::skip]
const C32_BYTE_MAP: [Option<u8>; 128] = [
    None,       None,       None,       None,       None,       None,       None,       None,
    None,       None,       None,       None,       None,       None,       None,       None,
    None,       None,       None,       None,       None,       None,       None,       None,
    None,       None,       None,       None,       None,       None,       None,       None,
    None,       None,       None,       None,       None,       None,       None,       None,
    None,       None,       None,       None,       None,       None,       None,       None,
    Some(0),    Some(1),    Some(2),    Some(3),    Some(4),    Some(5),    Some(6),    Some(7),
    Some(8),    Some(9),    None,       None,       None,       None,       None,       None,
    None,       Some(10),   Some(11),   Some(12),   Some(13),   Some(14),   Some(15),   Some(16),
    Some(17),   Some(1),    Some(18),   Some(19),   Some(1),    Some(20),   Some(21),   Some(0),
    Some(22),   Some(23),   Some(24),   Some(25),   Some(26),   None,       Some(27),   Some(28),
    Some(29),   Some(30),   Some(31),   None,       None,       None,       None,       None,
    None,       Some(10),   Some(11),   Some(12),   Some(13),   Some(14),   Some(15),   Some(16),
    Some(17),   Some(1),    Some(18),   Some(19),   Some(1),    Some(20),   Some(21),   Some(0),
    Some(22),   Some(23),   Some(24),   Some(25),   Some(26),   None,       Some(27),   Some(28),
    Some(29),   Some(30),   Some(31),   None,       None,       None,       None,       None,
];
```

stacks-core/src/crypto/hash.rs
stjepangolemac: I've now replaced all occurences with `<slice>.try_into()`. It's suggested as the best way to do this and the compiler will probably be able to optimize it as much as possible.

stjepangolemac: The only way to get rid of copying is to change types to borrow data, which might be interesting but potentially complicated. I don't think we need to care that much about performance.

stacks-core/src/lib.rs
stjepangolemac: From my experience preludes can be very handy when working with libraries. It decreases friction even further for beginners.

Let me know if you want to veto it and I'll remove it - no strong preferences from me.

stjepangolemac:
I've now also fixed all the tests 🎉

kantai:
```
c32 primitives
crypto primitives
basic types as principal data and hashes
StacksAddress
deposit wireformat and parsing logic
tests
```

Having a second rust implementation of all of this seems like the wrong approach to me. I would vastly prefer:

1. Identify what parts of the stacks-blockchain codebase you want to use in your library.
2. Move those into new (if necessary) workspace members (i.e., separating them from the main library output of the blockchain repo)
3. Trying to get the process started for altering the license on those.

Per the discussions in the last blockchain sync meeting about the licenses, it seemed like the folks involved at the Foundation were supportive of using more permissive licenses for parts of the codebase like this. Doing that now avoids a situation where there's multiple implementations of all of these things, requiring multiple audits, multiple testing frameworks, multiple reviews, and multiple maintenance for any and all required updates.

AshtonStephens:
stacks-core/address.rs abd stacks-core/c32.rs are the only files with unit tests - can you add unit tests to the rest of the files? I'd consider that part of the initial project structure. LMK if you agree others - in my experience it'll save time in the medium run

jcnelson:
Strong agree with @kantai.  All of the code here already lives in `stacks-network/stacks-blockchain`, so from a correctness and maintenance cost standpoint, it would make more sense to create a separate workspace for your library code over there.  Let me worry about the relicensing part; that shouldn't block you from submitting PRs there and getting them merged (since per the CLA, we can relicense it after-the-fact).

stjepangolemac:
I'd actually prefer extracting parts of `blockstack_lib` to a workspace member in the `stacks-blockchain` repo. The reason I didn't do it is to avoid licensing issues while I set this repo up.

@jcnelson are you saying I can create a new workspace member there with an MIT license or should I start with GPL and change later? Should I even put a LICENSE file in the workspace member subdirectory?

Also, @kantai I ultimately want to cover everything with tests. The reason not everything in the PR is tested is to keep the PR as small as possible. That said, I now have a clear idea of which parts of `blockstack_lib` I need to extract and can be more precise with PRs.

I can move this work to `stacks-blockchain` as soon as @jcnelson clarifies the path forward in terms of licensing.

netrome:
> That sounds good. I'm not sure if we can use that extracted code in sbtc-core before it's relicensed. Maybe we can leave this repo as is until the work in stacks-blockchain is complete.

Sounds like a good plan forward. And yeah, we can't link the extracted code here until it's relicensed. Also, the `stacks-core` code here could be broken out to a separate repo and be linked in `stacks-blockchain` - giving us a single source of truth for these data types.

stjepangolemac:
> Also, the stacks-core code here could be broken out to a separate repo and be linked in stacks-blockchain - giving us a single source of truth for these data types.

I believe what was suggested above is `stacks-core` being a package in the `stacks-blockchain` workspace, then later being relicensed into MIT.

netrome:
> I believe what was suggested above is stacks-core being a package in the stacks-blockchain workspace, then later being relicensed into MIT.

Right, but how would we depend on it in the sbtc-sdk repo before it's relicensed?

jferrant (approved):

stjepangolemac:
> Right, but how would we depend on it in the sbtc-sdk repo before it's relicensed?

We discussed this in the meantime and, for the record, we're unsure.

stjepangolemac:
Used all improvements I came across when trying to extract code from `blockstack_lib`. I'll abstain from asking for more reviews.

PR #8, opened 2023-07-27
StacksAddress construction improvements

stjepangolemac:
Removes `from_public_keys()` and exposes 4 distinct constructors. This simplifies the code and allows us to remove the hash mode enum. Also, argument types have been improved where possible.


jcnelson:
Hey! Thanks for taking a crack at this!

Some things to consider if you're going to merge this to the blockchain
repo:

* StacksAddress does not support p2wpkh or p2sh; it supports p2sh-p2wpkh
and p2sh-p2wsh. This information should be communicated by the new function
names. It was previously communicated in the docs for AddressHashMode.
There are no plans to add p2wsh or p2wsh, because StacksAddress is only
meant to encode Bitcoin address information for addresses that were
supported prior to Stacks 2.0 (StacksAddress inherits these hash modes in
order to provide backwards compatibility)

* The AddressHashMode enum needs to stay, because its values are stored in
the sortition DB as part of PoxAddress. Without AddressHashMode, this
information would not be able to be loaded (leading to a node crash).

These would need to be addressed before getting my approval in merging them
to stacks-blockchain. But since this is not stacks-blockchain, feel free to
do whatever :)

stjepangolemac:
Replaced by https://github.com/stacks-network/sbtc/pull/15

PR #10, opened 2023-07-27
Add logic and utilities for constructing sBTC transactions for Bitcoin

stjepangolemac:
- added construction logic for sBTC transactions
- reogranized the module structure so it's clearer where to import things from


sbtc-core/src/operations/construction/deposit.rs
```diff
@@ -0,0 +1,148 @@
...
+}
+
+fn deposit_data(recipient: &StacksAddress, network: Network) -> Vec<u8> {
+    magic_bytes(network)
+        .into_iter()
+        .chain(once(b'<'))
+        .chain(once(recipient.version() as u8))
+        .chain(recipient.hash().as_ref().to_owned())
+        .collect()
+}
```
netrome, line 148: Note: As we're moving beyond alpha, we want to have the encoding here be consistent with how a Clarity Principal is encoded according to `SIP-005`. This also ensures that the recipient can be a smart contract. We can extend this later, but just remember to create a ticket for it - as I assume you intend to do for the withdrawal_fulfillment and withdrawal_request.

netrome (approved):
Good stuff. Just a comment on changing wire formats for deposits.

stjepangolemac:
Replaced by https://github.com/stacks-network/sbtc/pull/17

sbtc-core/src/operations/construction/deposit.rs
stjepangolemac: Moved discussion here https://github.com/stacks-network/sbtc/issues/18

PR #17, opened 2023-08-03
Add logic and utilities for constructing sBTC transactions for Bitcoin

stjepangolemac:
Replaces https://github.com/stacks-network/sbtc/pull/10

- added construction logic for sBTC transactions
- reogranized the module structure so it's clearer where to import things from


stjepangolemac:
This is unchanged compared to the previous PR, which was approved. Will merge.

PR #27, opened 2023-08-06
Add signer scaffolding for signer and coordinator traits

jferrant:
This PR is a bit of a scaffolding piece for what I forsee the signer looking like. Please give feedback about whether you think its stupid to abstract out the coordinator and signer implementation. I thought this might work well for plugging FROST vs ROAST vs FIRE in there.


sbtc-core/src/signer/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+
+use crate::signer::config::Config;
+use crate::SBTCError;
+use bitcoin::{Network, PrivateKey, PublicKey, Transaction as BitcoinTransaction};
+use p256k1::ecdsa;
+use url::Url;
+use wsts::{bip340::SchnorrProof, common::Signature};
+
+#[derive(Default, Clone, Debug)]
+pub struct SignersData {
```
netrome, line 14: Name suggestion: How about just calling this "Signers"?

```suggestion
pub struct Signers {
```

sbtc-core/src/signer/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub signers: HashMap<u32, ecdsa::PublicKey>,
+    pub vote_shares: HashMap<u32, ecdsa::PublicKey>,
```
netrome, line 16: What's the difference between signers and vote shares here?

sbtc-core/src/signer/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+pub trait CoordinatorAlgorithm {
+    fn generate_sbtc_wallet_public_key(&self) -> Result<PublicKey, SBTCError>;
+    fn run_signing_round(
+        &self,
+        _tx: BitcoinTransaction,
+    ) -> Result<(Signature, SchnorrProof), SBTCError>;
+}
+
+/// Signer algorithm trait
+pub trait SignerAlgorithm: CoordinatorAlgorithm {
```
netrome, line 29: Why does a signer algorithm also have to be a coordinator algorithm?

sbtc-core/src/signer/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+            network,
+            stacks_node_rpc_url,
+            algorithm,
+        }
+    }
+
+    /// Get the current coordinator public key from the stacks node
+    pub fn get_coordinator_public_key(&self) -> Result<PublicKey, SBTCError> {
+        todo!()
+    }
```
netrome, line 73: What is the coordinator public key in this context? I was under the impression that the "coordinator" would be a different signer for each transaction to be signed.

sbtc-core/src/signer/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    /// Retrieve the current signers' data from the stacks node
+    pub fn get_signers_data(&self) -> Result<SignersData, SBTCError> {
+        todo!()
+    }
+
```
netrome, line 79: Nit: I wouldn't specify how the data is obtained here, only require that the data is available. In practice I assume this data is not fetched from the node on every request, but rather stored in an internal database (or in memory) for the signer, which subscribes to updates from the node to keep this data in sync.

netrome (approved):
Nice start, I have some questions - but see no reason to block merging this, hence approved!

sbtc-core/src/signer/mod.rs
stjepangolemac: Maybe this should map keys to vote shares (vice versa)?

sbtc-core/src/signer/mod.rs
stjepangolemac: I also believe if there's no reason to split these two I'd rather keep them in a single trait. Also, can't say I know Rust guidelines for trait names too well but something like `Coordinate` and `Sign` for some reason sounds better. A very small preference for this so please disregard if you don't agree.

sbtc-core/src/signer/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+    fn generate_sbtc_wallet_public_key(&self) -> Result<PublicKey, SBTCError>;
+    fn run_signing_round(
+        &self,
+        _tx: BitcoinTransaction,
+    ) -> Result<(Signature, SchnorrProof), SBTCError>;
+}
+
+/// Signer algorithm trait
+pub trait SignerAlgorithm: CoordinatorAlgorithm {
+    fn sign_message(&self, message: &[u8]) -> Result<Vec<u8>, SBTCError>;
```
stjepangolemac, line 30: Accepting a mutable buffer would prevent any unnecessary copies, but might not be viable if size changes dramatically.

`fn sign_message(&self, message: &mut [u8]) -> SBTCResult<()>`

sbtc-core/src/signer/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+    fn verify_message(
+        &self,
+        public_key: &ecdsa::PublicKey,
+        message: &[u8],
+    ) -> Result<bool, SBTCError>;
```
stjepangolemac, line 35: I suggest using `SBTCResult<T>`.

sbtc-core/src/signer/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+impl<S: SignerAlgorithm> Signer<S> {
+    /// Create a new signer
+    pub fn new(
+        config: Config,
+        private_key: PrivateKey,
+        network: Network,
+        stacks_node_rpc_url: Url,
+        algorithm: S,
```
stjepangolemac, line 59: It might be more idiomatic to pass `S` as a generic.

```
pub fn new<S>(...) -> Self;

...

let signer: SignerAlgorithm<MegaSigner> = ...;
```

sbtc-core/src/signer/mod.rs
stjepangolemac: Yeah, from experience getters should be implemented per trait implementor and then higher level functions can build default shared functionality on top of them.

stjepangolemac (approved):
Generally looks okay. I'd like to see some of the feedback addressed but we're still early so there's nothing wrong in moving fast and breaking things (builds).

sbtc-core/src/signer/mod.rs
jferrant: Signers maps signer ids to their public keys. Vote shares map key ids to public keys.

sbtc-core/src/signer/mod.rs
jferrant: Every signer also needs to be able to operate as a coordinator. But the logic behind coordination is independent of the logic for the actual math behind the signing.

sbtc-core/src/signer/mod.rs
jferrant: You would want to know if the person requesting you to sign is the valid coordinator for the given round or potentially transaction. I was not convinced that it being per transaction makes sense.

sbtc-core/src/signer/mod.rs
jferrant: This would most likely be stored in the stacker db which is accessed via a stacks node.

sbtc-core/src/signer/mod.rs
jferrant: I don't think accepting a mutable buffer here would help much given the sign_msg functions that seem to be used throughout the crypt side where we use a hash function to create a mutable nbuffer and return that. Wouldbe hard to avoid this copy.

netrome: Why would the message ever be mutated?

sbtc-core/src/signer/mod.rs
jferrant: Not sure how to do this when its already a generic stored as a field in the struct....

sbtc-core/src/signer/mod.rs
stjepangolemac: Ah okay got it. In some sense both coordinating and signing is a part of the same process, but maybe it will be easier to understand and present if separated into two distinct traits?

sbtc-core/src/signer/mod.rs
stjepangolemac: I didn't mean mutating the message but from the bdk it seems to me like it keeps an unsigned transaction under the hood, and just modifies it when it signs it. That's why I thought a buffer might come in handy here.

netrome: Ah yeah I guess the low-level signer API is about creating a signature for arbitrary bytes. A high-level API that adds a signature to a transaction could mimic this.

sbtc-core/src/signer/mod.rs
jferrant: Oh I am misunderstanding your feedback. Yes good point, they should be sep traits, not inheriting from each other. Fixed!

sbtc-core/src/signer/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+use crate::SBTCError;
+use crate::{signer::config::Config, SBTCResult};
+use bitcoin::{Network, PrivateKey, PublicKey, Transaction as BitcoinTransaction};
+use p256k1::ecdsa;
+use url::Url;
+use wsts::{bip340::SchnorrProof, common::Signature};
+
+#[derive(Default, Clone, Debug)]
+/// Signers' data required for weighted distributed signing
+pub struct Signers {
```
xoloki, line 15: I don't like the name `Signers` here.  The encapsulated data is the public keys for all signers, indexed by their signer IDs and key IDs, not the signers themselves.

I think we called this `SignerKeys` or `SignerPublicKeys` previously?  Since we're in the `signers` module here I'd probably just call it `PublicKeys`, and skip the excess `Signer` token.

sbtc-core/src/signer/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+    /// Signer id to public key mapping
+    pub signer_ids: HashMap<u32, ecdsa::PublicKey>,
+    /// Vote share to public key mapping
+    pub vote_shares: HashMap<u32, ecdsa::PublicKey>,
+}
+
+/// sBTC Signer trait
+pub trait SBTCSigner {
```
xoloki, line 23: This trait is again just returning public keys, so the name doesn't seem particularly descriptive.  So I'd probably call the previous trait `SignerPublicKeys`, and this trait just `PublicKeys` (or `SBTCPublicKeys` if you really want the extra token).

sbtc-core/src/signer/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+    fn signers(&self) -> SBTCResult<Signers>;
```
xoloki, line 24: And I'd call this function `signer_public_keys` so it's clear what's being returned.

xoloki (changes requested):
I think the structs/traits/fns that deal with public keys need to encode that into their names.  But the basic layering of the traits seems correct.

sbtc-core/src/signer/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+    fn coordinator_public_key(&self, _tx: BitcoinTransaction) -> SBTCResult<PublicKey>;
```
xoloki, line 25: Alternately, if you put the token `PublicKeys` into the trait name, then we could drop `_public_key` from the fn names.

xoloki (approved):
LGTM thanks!

sbtc-core/src/signer/mod.rs
netrome: Sure, I'd imagine that being something that the signer can derive from the information it has. I.e. if it knows the set of signers and which round or tx it regards, it would be able to easily compute who is the coordinator (primary and secondary).

sbtc-core/src/signer/mod.rs
stjepangolemac: Actually, ignore what I wrote for this one. When a struct has a generic type I usually expose it as a generic parameter of the function, not as an argument. If both works then it's just a matter of preference and I don't really have a strong one here.

sbtc-core/src/signer/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// Signer ids to public key mapping
+    pub signer_ids: HashMap<u32, ecdsa::PublicKey>,
+    /// Vote ids to public key mapping
+    pub vote_ids: HashMap<u32, ecdsa::PublicKey>,
+}
+
+/// sBTC Keys trait for retrieving signer IDs, vote IDs, and public keys
+trait Keys {
+    fn public_keys(&self) -> SBTCResult<PublicKeys>;
+    fn coordinator_public_key(&self, tx: &BitcoinTransaction) -> SBTCResult<PublicKey>;
```
netrome, line 35: Should we have this method return an ordered list of public keys, the first one being the primary coordinator, second one being the secondary etc.?

sbtc-core/src/signer/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+        tx: &BitcoinTransaction,
+    ) -> SBTCResult<(Signature, SchnorrProof)>;
+}
+
+/// Sign trait for signing and verifying messages
+pub trait Sign {
+    fn sign_message(&self, message: &[u8]) -> SBTCResult<Vec<u8>>;
+    fn verify_message(&self, public_key: &ecdsa::PublicKey, message: &[u8]) -> SBTCResult<bool>;
+}
+
```
netrome, line 53: A trait we can add later is also "Reveal" to reveal committed bitcoin transactions. I'll create a follow-up ticket for that.

netrome: Ticket: https://github.com/stacks-network/sbtc/issues/30

netrome (approved):

PR #29, opened 2023-08-07
Introduce CI with xtask script

AshtonStephens:
## Summary

*Note: This is a draft pull request*

This adds an `xtask` rust cli to the page and then utilizes that script within github workflows. This is the `--help` print from the script:

```
DevOps Script for local debugging also used in github workflows

Usage: xtask [OPTIONS] <COMMAND>

Commands:
  workflow, -w  Run a workflow comprised of a sequence of steps
  steps, -s     Run a sequence of steps in the order that they're written
  install, -i   Install all components and crates used by the xtask script
  clean, -c     Clean the directory of all build artifacts
  help          Print this message or the help of the given subcommand(s)

Options:
  -l, --lazy-install
          Whether to install dependencies lazily.
          
          This flag is especially helpful for github workflows, in which we don't want to install more dependencies than we need to for the specific workflow.

  -m, --min-coverage <MIN_COVERAGE>
          Min Coverage
          
          This only makes sense when the commands includes the step `verify-coverage-percent`, otherwise it's unused.
          
          [default: 50]

  -h, --help
          Print help (see a summary with '-h')
```

**Disclaimer:** This is my first rust program so let me know if I'm doing anything wrong.

This Pr includes a rough draft README.md within the xtask cli package as well.


.github/workflows/rust.yml
```diff
@@ -0,0 +1,44 @@
+name: Rust
+
+on:
+
+  pull_request:
+    branches: [ "master", "20-introduce-ci" ] # We should make this main or mainline.
+    paths-ignore: ['**.md']
```
wileyj, line 7: definitely would add an ignore for yml (possibly json, since i see a changed json file in this PR)

wileyj: it might be better to trigger only on *code* changes though, toml or rs only

sergey-shandar (commented):
please, remove `.vscode/settings.json`.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "livePreview.defaultPreviewPath": "/coverage/html/index.html"
```
stjepangolemac, line 2: Let's remove and gitignore this file.

Cargo.toml
```diff
@@ -1,5 +1,10 @@
 [workspace]
-members = ["sbtc-core", "stacks-core"]
+members = [
+    "sbtc-core",
+    "stacks-core",
+    "xtask",
+]
+resolver = "2"
```
stjepangolemac, line 7: It's possible this field is unnecessary according to this page:
https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html

xtask/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+edition = "2021"
+name = "xtask"
+version = "0.1.0"
+
+[dependencies]
+clap = { version="4.3", features = ["derive"]}
+clap-verbosity-flag = "2.0.1"
+anyhow = "1.0"
+duct = "0.13"
+colored = "2"
```
stjepangolemac, line 11: Should these dependencies be workspace dependencies or not? I honestly don't know. What do you think @AshtonStephens @wileyj?

stjepangolemac (approved):
Looks good 🎉

AshtonStephens:
Removing this draft PR because we've opted to go with a different taskrunner - cargo make: https://github.com/stacks-network/sbtc/pull/52

After comments in the task runner discussion we'll likely finally switch to classic make: https://github.com/orgs/stacks-network/discussions/491

PR #35, opened 2023-08-08
Add auto deny no docs for public functions throughout stacks-core

jferrant:


AshtonStephens (approved):
Looks good. Why not use `#[forbid(missing_docs)]`?

netrome (approved):

jferrant:
> Looks good. Why not use `#[forbid(missing_docs)]`?

I did the deny option originally because was unsure if we would ever want to override this feature. Pretty sure if you do forbid, it will prevent any other type of the lint being present and I thought we might end up moving over code from stacks blockchain and use "#[warn(missing_docs]" as placeholder for the more appropriate dev to fill out the documentation

However, if people think this will never really be a use case , can change to forbid. Alternatively, we could always change back to deny if we find forbid too restrictive.

AshtonStephens (approved):
> Pretty sure if you do forbid, it will prevent any other type of the lint being present and I thought we might end up moving over code from stacks blockchain and use "#[warn(missing_docs]" as placeholder for the more appropriate dev to fill out the documentation

Yes, that's what I think too. But imo it's better to make it hard to overwrite so we have a conversation about if when we choose to do it. 

Looks good!

PR #37, opened 2023-08-08
Add revealer trait

jferrant:


xoloki (approved):
Lgtm

sbtc-core/src/signer/mod.rs
```diff
@@ -57,6 +57,12 @@ pub trait Sign {
     fn verify_message(&self, public_key: &ecdsa::PublicKey, message: &[u8]) -> SBTCResult<bool>;
 }
 
+/// Revealer trait for revealing BTC commit transactions
+trait Revealer {
+    /// Create a reveal transaction from the BTC commit transaction
+    fn reveal_transaction(&self, tx: &BitcoinTransaction) -> SBTCResult<BitcoinTransaction>;
+}
+
```
netrome, line 65: We should add at least a placeholder type here to represent the taproot internal key and the merkle tree of scripts, which is needed to create the reveal transaction. This type could potentially be used: https://docs.rs/bitcoin/0.30.1/bitcoin/taproot/struct.TaprootSpendInfo.html

netrome (changes requested):
We need more information to reveal the commit transaction, which should be reflected in the trait.

sbtc-core/src/signer/mod.rs
jferrant: Yes let me give the doc a better read over so I can better reflect in the trait.

jferrant: the more I think about this, the more I don't think this belongs in the signer. Signer's should always be able to deny or approve a transaction and they should be governing the protocol itself. They should only be searching for the already revealed transaction. Not doing the logic to create it especially as this may mean that a signer has to be responsible for returning the funds to the initial committer( if the sBTC tx gets denied )which violates the idea that they may not want to interact with this entity. This should be an agreement between the committer and revealer, a one to one relationship, not a one to many essentially.

netrome: Yeah we need to sync again. I agree with the point that I would like to have the Signer and Revealer be separate entities.

However, enforcing the signers to also be the Revealer does allow the signers to deny requests by simply ignoring commit transactions. Since commit transactions are invisible on-chain, no one knows that they are payments to the signers until they are revealed. If they are never revealed, the corresponding BTC will never be deposited. This is the mechanism we're piggy backing on to use commit-reveal to allow signers to deny transactions.

But won't ignoring commit transactions lock up the BTC forever? No, since the commit transaction may commit to multiple spend paths - the person creating the commit transaction can have a reclaim spend path allowing them to take back their BTC if the reveal tx never happens.

I'd love to chat more of this. I feel this is part of the design that we have taken for granted for too long, and there are probably better ways to do it than what's currently on the table.

stjepangolemac (approved):
It looks good, let's replace these enums with the ones from the sbtc-core when they're ready. Thanks!

sbtc-core/src/signer/coordinator/mod.rs
```diff
@@ -25,8 +27,41 @@ pub struct PublicKeys {
...
+/// depending on https://github.com/Trust-Machines/stacks-sbtc/pull/595
+pub enum SBTCTransaction {
+    /// A commit Bitcoin transaction
+    Commit(TaprootSpendInfo, BitcoinTransaction),
+    /// A withdrawal Stacks transaction
+    Withdawal(StacksTransaction),
+}
+
+/// Revealer trait for revealing BTC commit transactions
+pub trait Revealer {
```
netrome, line 42: Should the trait be called "Reveal" to be consistent with the "Sign" and "Coordinate" traits?

sbtc-core/src/signer/mod.rs
```diff
@@ -39,6 +45,12 @@ pub trait Sign {
...
 }
 
+/// Validator trait for validating pending Bitcoin transactions
+pub trait Validator {
+    /// Validate the given signable Bitcoin transaction
+    fn validate_transaction(&self, tx: &SignableTransaction) -> SBTCResult<bool>;
+}
+
 /// sBTC compliant Signer
 pub struct Signer<S: Sign + Coordinate> {
```
netrome, line 55: Although unrelated to this PR, I realized it's bad practice to add trait bounds to structs - since structs are about data organization, while traits are about behavior. Best practices is to just omit the trait bounds in the struct definition, and use them in the impl blocks.

```suggestion
pub struct Signer<S> {
```

sbtc-core/src/signer/mod.rs
```diff
@@ -125,3 +132,22 @@ impl<S: Sign + Coordinate> Keys for Signer<S> {
         todo!()
     }
 }
+
+impl<S: Sign + Coordinate> Validator for Signer<S> {
```
netrome, line 136: I assume we'll need `Reveal` here too?
```suggestion
impl<S: Sign + Coordinate + Reveal> Validator for Signer<S> {
```

sbtc-core/src/signer/coordinator/mod.rs
```diff
@@ -25,8 +27,41 @@ pub struct PublicKeys {
...
+    /// Retrieve pending Commit transactions information
+    fn pending_commit_transactions(
+        &self,
+    ) -> SBTCResult<Vec<(TaprootSpendInfo, BitcoinTransaction)>>;
+    /// Validate the given commit transaction
```
netrome, line 47: This implies the state lives in the Revealer. I wonder if it would be better if we managed state somewhere else in a single place to ensure database consistency. The Revealer trait would only revolve around the functionality to validate commit transactions and reveal transactions. The process of interacting with a reveal server (or however we'll implement it) could be broken out to a separate entity.

netrome (approved):
Some thoughts, but generally looks good to me

sbtc-core/src/signer/coordinator/mod.rs
```diff
@@ -25,8 +27,41 @@ pub struct PublicKeys {
...
+        &self,
+        spend_info: TaprootSpendInfo,
+        tx: &BitcoinTransaction,
+    ) -> SBTCResult<BitcoinTransaction>;
+}
+
 /// Coordinator trait for generating the sBTC wallet public key and running a signing round
 pub trait Coordinate {
+    /// Retrieve pending sBTC transactions to be processed
+    fn pending_transactions(&self) -> SBTCResult<Vec<SBTCTransaction>>;
```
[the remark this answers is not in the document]
jferrant: This is to retrieve transactions from the stacks node. Not the database state.

sbtc-core/src/signer/mod.rs
jferrant: I did it this way to ensure if a user wants to create their own signer using the SDK, they cannot without implementing these traits. But can def pull it out.

sbtc-core/src/signer/coordinator/mod.rs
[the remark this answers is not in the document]
jferrant: I will change to sbtc_transactions to make it clear its just to retrieve from the blockchain, not the database and no state is implied from it. The outer signer would handle the results and its state.

sbtc-core/src/signer/coordinator/mod.rs
```diff
@@ -25,8 +27,39 @@ pub struct PublicKeys {
 /// TODO: Define the Message types for Tx Signning Round
 /// https://github.com/stacks-network/sbtc/issues/43
 
+/// An sBTC transaction needing to be processed by the coordinator
+/// TODO: replace with the core library's SBTCTransaction
+/// This could be a BTC transaction or a STX transaction
+/// depending on https://github.com/Trust-Machines/stacks-sbtc/pull/595
+pub enum SBTCTransaction {
+    /// A commit Bitcoin transaction
+    Commit(TaprootSpendInfo, BitcoinTransaction),
```
xoloki, line 36: You can't construct a `TaprootSpendInfo` for a script spend unless you know all the leaves of the script tree.  My understanding of commit/reveal is that we only know how to spend one of the paths.

We can get around this if we specify that the reveal is actually an internal key spend with the aggregate key as the internal key.

xoloki (approved):
Looks fine, but I'm still having issues with commit/reveal, it seems like we're hand-waving away some of the critical bits @netrome @jferrant

sbtc-core/src/signer/coordinator/mod.rs
xoloki: If we aren't going to use an internal key spend then we need a `ControlBlock` not a `TaprootSpendInfo`.

PR #39, opened 2023-08-08
Add coordinator scaffolding

jferrant:


sbtc-core/src/signer/mod.rs
```diff
@@ -143,31 +125,3 @@ impl<S: Sign + Coordinate> Keys for Signer<S> {
         todo!()
     }
 }
-
```
xoloki, line 146: Did you mean to remove the skeleton implementations?

xoloki (approved):

sbtc-core/src/signer/coordinator/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+/// Signers' public keys required for weighted distributed signing
+pub struct PublicKeys {
+    /// Signer ids to public key mapping
+    pub signer_ids: HashMap<u32, ecdsa::PublicKey>,
+    /// Vote ids to public key mapping
+    pub vote_ids: HashMap<u32, ecdsa::PublicKey>,
+}
+
+/// TODO: Define the Message types for DKG round
+/// TODO: Define the Message types for Tx Signning Round
```
netrome, line 23: Bonus points for linking issues in the todo comments

netrome (approved):

sbtc-core/src/signer/mod.rs
jferrant: Yes I did. The coordinator should actually be implemented on the frost-coordinator, fire-coordinator, roast-coordinator level and then passed to the Signer struct.

sbtc-core/src/signer/coordinator/mod.rs
jferrant: Done

Issue #18, opened 2023-08-03
Principal encoding for sBTC transactions


stjepangolemac:
Quoting @netrome's [comment](https://github.com/stacks-network/sbtc/pull/10#discussion_r1279009359):

> As we're moving beyond alpha, we want to have the encoding here be consistent with how a Clarity Principal is encoded according to SIP-005. This also ensures that the recipient can be a smart contract.

I've skimmed through the [SIP-005](https://github.com/stacksgov/sips/blob/main/sips/sip-005/sip-005-blocks-and-transactions.md) now but couldn't find the difference between our address encoding and it. Maybe you meant we need to be able to encode both contracts and ordinary addresses here?

I'll assign you on this one but feel free to remove yourself and just leave a comment.

netrome:
Are we adding the `0x05` or `0x06` prefix indicating that the principal is a contract address already? 
![image](https://github.com/stacks-network/sbtc/assets/22524681/65fdb426-8aa8-4a97-99f8-f730a0707797)

Otherwise IMO that's the only difference.

stjepangolemac:
Yeah we're not doing that at the moment. Thanks for the clarification.

stjepangolemac:
Closed by https://github.com/stacks-network/sbtc/pull/41

PR #41, opened 2023-08-09
Add length prefixing to contract names in deposits

stjepangolemac:
Adds length prefixed contract names in deposit data.

Addresses all issues from the original PR:

- https://github.com/stacks-network/stacks-blockchain/pull/3718/files#r1258665342
- https://github.com/stacks-network/stacks-blockchain/pull/3718/files#r1258666082
- https://github.com/stacks-network/stacks-blockchain/pull/3718/files#r1258668073

There are a few more things I'd like to check before considering this ready:

- [x] Do we need reproducible tests considering they generate random data? (a bit of extra work but nothing major)
- [x] Now with the contract names being length prefixed do we want to reserve all remaining bytes for memo or keep it restricted to the current range?
- [x] Using correct principal encoding? https://github.com/stacks-network/sbtc/issues/18

Another thing I've done in this PR is changed the parsing logic to include full 80 bytes from the Bitcoin transaction output. I think it makes slicing of the data clearer. Let me know what you think.

### DIFF
--- a/sbtc-core/src/operations/parsing/deposit.rs
+++ b/sbtc-core/src/operations/parsing/deposit.rs
@@ -1,29 +1,52 @@
 /*!
-Deposit is a transaction with the output structure as below:
+Deposit is a Bitcoin transaction with the output structure as below:
 
 1. data output
 2. payment to peg wallet address
 
-The data output should contain data in the following format:
+The data output should contain data in the following byte format:
 
-0      2  3                  24                            65       80
-|------|--|------------------|-----------------------------|--------|
- magic  op   Stacks address    Size prefixed contract name    memo
-                                   (optional, see note)
+```text
+0     2  3                                                                    80
+|-----|--|---------------------------------------------------------------------|
+ magic op                           deposit data
+```
 
-Contract name is prefixed by a single byte size followed by the contract name
-bytes.
+Where deposit data should be in the following format:
+
+```text
+3                                                      25 >= N <= 66          80
+|------------------------------------------------------------------|-----------|
+                           principal data                              extra
+                                                                       bytes
+```
+
+There are two types of principal data:
+
+- standard (includes only principal type, address version and address hash)
+- contract (includes everything from the last diagram)
+
+If the principal data is of the contract type, then the contract name cannot be
+longer than 40 characters.
+
+Principal data should be in the following format:
+
+```text
+3         4         5                25       26                         N <= 66
+|---------|---------|-----------------|--------|-------------------------------|
+ principal  address       address      contract            contract
+   type     version        hash          name                name
+                                      length (N)
 ```
 */
-use std::io::Cursor;
+use std::io::{Cursor, Read};
 
-use stacks_core::{
-    address::{AddressVersion, StacksAddress},
-    codec::Codec,
-    utils::{ContractName, PrincipalData, StandardPrincipalData},
-};
+use stacks_core::{codec::Codec, utils::PrincipalData, StacksError};
 
 use crate::{SBTCError, SBTCResult};
+
+const WIRE_DATA_MIN_LENGTH: usize = 25;
+const WIRE_PRINCIPAL_DATA_INDEX: usize = 3;
 
 /// Contains the parsed data from a deposit transaction
 pub struct ParsedDepositData {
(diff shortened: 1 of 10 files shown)


netrome:
> There are a few more things I'd like to check before considering this ready:
> 
>     * [ ]  Do we need reproducible tests considering they generate random data? (a bit of extra work but nothing major)
> 
>     * [ ]  Now with the contract names being length prefixed do we want to reserve all remaining bytes for memo or keep it restricted to the current range?

1. Reproducible tests, yes I'd like to have that. Is there anything needed beyond seeding the rng?

2. In this context we shouldn't even consider the memo as I see it. That's just bytes that Stacks miners opportunistically use for communication. It's not part of the sBTC spec. Therefore, there's no memo mentioned in the sBTC docs either https://stacks-network.github.io/sbtc-docs/sbtc-operations/bitcoin-transactions.html

stacks-core/src/codec.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    fn deserialize<R: io::Read>(data: &mut R) -> StacksResult<Self>
+    where
+        Self: Sized,
+    {
+        Self::codec_deserialize(data).map_err(|err| CodecError::IoError(err).into())
+    }
+
+    fn serialize_to_vec(&self) -> Vec<u8> {
+        let mut buffer = Vec::with_capacity(21);
```
netrome, line 34: Having a default capacity of 21 feels very weird here.

stacks-core/src/codec.rs
```diff
@@ -0,0 +1,40 @@
...
+        self.codec_serialize(dest)
+            .map_err(|err| CodecError::IoError(err).into())
+    }
+
+    fn deserialize<R: io::Read>(data: &mut R) -> StacksResult<Self>
+    where
+        Self: Sized,
+    {
+        Self::codec_deserialize(data).map_err(|err| CodecError::IoError(err).into())
+    }
```
netrome, line 31: I don't feel like these methods really add anything. I'd prefer to have a provided serialize function outside the trait for specific return types, similar to how the bitcoin crate does it https://docs.rs/bitcoin/0.30.1/bitcoin/consensus/encode/fn.serialize.html

netrome (changes requested):
The code generally looks good. I just would like to have the tests reproducible as previously discussed. Also, I think this would be a good opportunity to also incorporate the type prefix to have the full principal encoding supported as per https://github.com/stacks-network/sbtc/issues/18

stacks-core/src/codec.rs
stjepangolemac: Oh that's a good catch, a leftover from the address serialization code.

stacks-core/src/codec.rs
stjepangolemac: I'm not sure I understand. That `serialize()` function from the `bitcoin` crate requires the type to implement `Encodable` which is basically the same as `Codec` here.

I could add a serialize function to the `stacks-core` if you think that would improve ergonomics?

stacks-core/src/codec.rs
```diff
@@ -0,0 +1,52 @@
+/*!
+Module for contract name parsing
```
jferrant, line 2: Is codec really for contract name parsing? I thought that was the contract_name.rs module. I assume this is a copy paste error.

stjepangolemac: Yep not sure how that happened. It's fixed now.

stacks-core/src/codec.rs
netrome: Yeah I agree about the required functions in the Codec trait. I was referring to the two provided methods here which just seems to wrap the `codec_serialize` and `codec_deserialize` functions with a new error type.

sbtc-core/src/operations/parsing/deposit.rs
```diff
@@ -1,80 +1,261 @@
...
+longer than 40 characters.
+
+Principal data should be in the following format:
+
+```text
+3         4         5                25       26                         N <= 66
+|---------|---------|-----------------|--------|-------------------------------|
+ principal  address       address      contract            contract
+   type     version        hash          name                name
+                                      length (N)
```
netrome, line 39: Nice!

sbtc-core/src/operations/parsing/deposit.rs
```diff
@@ -1,80 +1,261 @@
...
-            .map_err(|_| {
-                SBTCError::MalformedData("Byte data is larger than 20 bytes for the address")
-            })?;
-
-        StandardPrincipalData::new(
-            version,
-            StacksAddress::new(version, Hash160Hasher::new(address_data)),
-        )
+    let mut data = Cursor::new(&data[WIRE_PRINCIPAL_DATA_INDEX..]);
+    let recipient = PrincipalData::deserialize(&mut data)?;
```
netrome, line 66: Looking great!

netrome (approved):
Great stuff!

stacks-core/src/codec.rs
```diff
@@ -0,0 +1,52 @@
...
+        Self: Sized,
+    {
+        Self::codec_deserialize(data).map_err(|err| CodecError::IoError(err).into())
+    }
+
+    /// Serialize to a vector
+    fn serialize_to_vec(&self) -> Vec<u8> {
+        let mut buffer = vec![];
+
+        self.serialize(&mut buffer).unwrap();
```
netrome, line 48: Btw is this unwrap safe? I'm sceptical towards it

stacks-core/src/codec.rs
stjepangolemac: Discussed this on Slack and decided to leave potential updates for future PRs.

stacks-core/src/codec.rs
stjepangolemac: Yeah I believe it is. It would probably fail only when there's not enough memory on the system to allocate for the vector. I think we can safely "ignore" that case as a lot of other things will start failing.

PR #44, opened 2023-08-10
Commit reveal transactions

stjepangolemac:
- moved all commit reveal logic we had in the old issue and branch https://github.com/Trust-Machines/stacks-sbtc/issues/281
- added documentation where needed

Some additional questions that I'm thinking about:

- [x] I suppose we also want commit reveal transaction parsing in this one?
- [x] Do you think the module structure is okay, operations module and everything in it?
- [x] It looks to me that we can replace tx output parsing with Codec implementations for types such as `DepositData`, what do you think?


netrome:
>  It looks to me that we can replace tx output parsing with Codec implementations for types such as DepositData, what do you think?

This sounds interesting. Spontaneously I like it!

jferrant:
> > It looks to me that we can replace tx output parsing with Codec implementations for types such as DepositData, what do you think?
> 
> This sounds interesting. Spontaneously I like it!

I like.

sbtc-core/src/operations/commit_reveal/construction.rs
```diff
@@ -0,0 +1,157 @@
...
+            .chain(empty_memo)
+            .chain(self.reveal_fee.to_be_bytes())
+            .collect())
+    }
+}
+
+/// Constructs a peg in payment address
+pub fn deposit_commit<'p>(
+    deposit_data: DepositData<'p>,
+    revealer_key: &XOnlyPublicKey,
```
jferrant, line 104: Does this revealer key refer to the revealer service or the signer who actually constructs the corresponding reveal BTC transaction? I assume the signer. which means that the user could dictate which signer triggers the reveal transaction ? This implies technically that anyone could construct a BTC transaction and sign it themselves and the signers potentially need to do more than just reveal transactions. They need to check for otehr revealed transactions that some other entity has revealed so they can validate it construct the corresponding stx transaction.

jferrant: In mini, this revealer_key would be the aggregate public key from the DKG rounds corresponding private key then?

xoloki: @jferrant and I discussed this and related issues today.  We came to a number of conclusions which we will want to discuss amongst the team tomorrow.  I'll put some of them down here for reference:

1. For commit/reveal deposits, the commit txo should be a taproot txo with two spend paths: a key spend using the sbtc wallet key, and a script spend so the sender can reclaim
2. The signers only need to know the merkle root of the script tree to be able to do a key spend, which minimizes what needs to be backchanneled
3. Commit txos can be trivially stolen if spendable by an external reveal service; the sbtc wallet however has a threshold signature with the reputation of the sbtc system to prevent fraud
4. For OP_RETURN deposits, if the deposit txo is a taproot txo then the OP_RETURN txo needs to communicate the merkle root in addition to the STX principal
5. OP_RETURN/taproot deposits should be spent during the signing process if the deposit is accepted to prevent reclaim after mint
6. OP_RETURN/p2pkh deposits don't need to be spent if deposit is accepted, but then sender will have no way to reclaim if deposit is rejected

Open questions:

1. If an OP_RETURN/p2pkh deposit is rejected, what is the best course of action legally?  Is it okay to return a deposit to a (possibly) prohibited individual?  Would it be better to burn the txo?  Is there a FINCEN/OFAC/etc place to surrender such txos?
2. If we can't return and/or shouldn't burn a rejected deposit, what do we do on wallet handoff?  Do we need to leave any such utxos in the old sbtc wallet forever?  

@netrome let's discuss this tomorrow

sbtc-core/src/operations/commit_reveal/construction.rs
```diff
@@ -0,0 +1,157 @@
...
+            address,
+            contract_name,
+            reveal_fee,
+        }
+    }
+
+    /// Serializes this data according to the SIP-021 wire formats
+    /// Links:
+    ///  - [SIP draft](https://github.com/stacksgov/sips/blob/56b73eada5ef1b72376f4a230949297b3edcc562/sips/sip-021/sip-021-trustless-two-way-peg-for-bitcoin.md)
+    ///  - [Reference implementation](https://github.com/stacks-network/stacks-blockchain/blob/next/src/chainstate/burn/operations/peg_in.rs)
```
netrome, line 44: SIP-021 is deprecated. I think it's better to point to the sBTC docs. This is the new reference implementation ;)

sbtc-core/src/operations/commit_reveal/construction.rs
netrome: Happy to discuss commit-reveal. Regarding some points:

2. The signers also need to know the internal public key to spend the script path, and the merkle path to the script.
3. Yes, committers should only work with revealers they trust. This was also noted in the original design doc:

> Security note: The Revealer has the possibility to steal the commit transaction for it's own use. Doing this is a visible crime on the chain, but it relies on the Revealer to be a trusted entity. If the Revealer is operated by Stackers, it is possible to implement mechanisms to penalize such behavior.

https://github.com/Trust-Machines/stacks-sbtc/blob/main/sbtc-planning/commit-reveal-ops.md#commit-reveal-protocol

4. I'm a bit confused here. Let's talk more.

5. I assume you mean OP_DROP?

6. Yes, that's true. That's why we don't support OP_RETURN for 0.1.

**Open questions**
1. I also wonder about this one. A related question, what happens if you just receive a random payment from a blacklisted address? Should you return it or burn it?

2. Agree, we need to discuss this. This could potentially require sBTC UX changes such as dropping OP_RETURN support for 1.0 - which I have learned some people are assuming that we do.

sbtc-core/src/operations/commit_reveal/utils.rs
```diff
@@ -0,0 +1,171 @@
...
+
+    BitcoinAddress::p2tr(
+        &secp,
+        spend_info.internal_key(),
+        spend_info.merkle_root(),
+        Network::Testnet, // TODO: Make sure to make this configurable
+    )
+}
+
+// TODO: Figure out if we want this to be public or not
```
netrome, line 89: Let's remove this todo and make the call
```suggestion
```

sbtc-core/src/operations/commit_reveal/utils.rs
```diff
@@ -0,0 +1,171 @@
...
+fn internal_key() -> UntweakedPublicKey {
+    // Copied from BIP-0341 at https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#constructing-and-spending-taproot-outputs
+    // The BIP recommends a point lift_x(0x0250929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0).
+    // This hex string is copied from the lift_x argument with the first byte stripped.
+
+    // TODO: Verify that this point is secure
+    let internal_key_vec =
+        hex::decode("50929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0").unwrap();
+
+    // TODO: Confirm that this is infallible
```
netrome, line 45: Since we have unit tests this is infallible
```suggestion
```

sbtc-core/src/operations/commit_reveal/utils.rs
```diff
@@ -0,0 +1,171 @@
...
+
+/// Commit reveal result
+pub type CommitRevealResult<T> = Result<T, CommitRevealError>;
+
+fn internal_key() -> UntweakedPublicKey {
+    // Copied from BIP-0341 at https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#constructing-and-spending-taproot-outputs
+    // The BIP recommends a point lift_x(0x0250929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0).
+    // This hex string is copied from the lift_x argument with the first byte stripped.
+
+    // TODO: Verify that this point is secure
```
netrome, line 41:
```suggestion
```

sbtc-core/src/operations/commit_reveal/utils.rs
netrome: Oh actually we don't. Should we perhaps add some unit tests?

netrome: Perhaps a big ask, but ideally we'd have a unit test spinning up a bitcoin regtest node, constructing and broadcasting a commit UTXO, and constructing a reveal for it.

netrome (dismissed):
Good stuff. Should we add tests as a follow-up?

sbtc-core/src/operations/commit_reveal/construction.rs
netrome: I sent and invite to explore this before our team meeting today. Let me know if the time doesn't work.

netrome: > In mini, this revealer_key would be the aggregate public key from the DKG round?

Yes

sbtc-core/src/operations/commit_reveal/construction.rs
stjepangolemac: Will do 👍

sbtc-core/src/operations/commit_reveal/utils.rs
stjepangolemac: Discussed the testing plan on Slack:

- We test parsing of transactions with some custom tx output generation code (similar to what I did for op return parsing module tests)
- We then write real construction logic for the same tx types and replace the custom tx output generation code - then use parsing logic to test these
- In some way construction logic would test the parsing logic, and vice versa
- We could then extend the testing situation by spinning a BTC node and trying to submit these transactions

sbtc-core/src/operations/commit_reveal/construction.rs
```diff
@@ -0,0 +1,151 @@
...
+    reclaim_key: &XOnlyPublicKey,
+) -> CommitRevealResult<BitcoinAddress> {
+    commit(
+        &withdrawal_data.serialize_to_vec(),
+        revealer_key,
+        reclaim_key,
+    )
+}
+
+/// Constructs a transaction that reveals the peg in payment address
```
jferrant, line 113: We might want to remove all references to peg in and peg out throughout the code even if its in comments.

sbtc-core/src/operations/commit_reveal/construction.rs
```diff
@@ -0,0 +1,151 @@
...
+        Ok(Self {
+            amount,
+            signature,
+            reveal_fee,
+        })
+    }
+}
+
+/// Constructs a peg in payment address
+pub fn deposit_commit(
```
jferrant, line 92: I do find it a bit confusing that deposit_commit() doesn't create a deposit_commit transaction. Maybe rename to deposit_commit_address()?

jferrant (approved):

sbtc-core/src/operations/commit_reveal/construction.rs
stjepangolemac: Yep you're right 👍

stjepangolemac:
@netrome yeah it might be better if we add tests in a separate PR to keep the ball rolling.

I've once again reorganized the operations module, I think this is the best structure since we started. All operations now have output data structs that can we can cleanly serialize and deserialize.

This PR has been approved 2 times already so I consider it ready for merge. In case you think of any additional things we can add them on Monday. If not, feel free to merge it first thing on Monday, or I will.

jferrant (approved):

stjepangolemac:
I see that clippy would want us to fix a few cosmetic things before merging but I'll merge either way. I don't want to ask you guys for another round of interviews. Thank you!

PR #49, opened 2023-08-12
Signer SDK: sep blockchain communication logic into sep broker

jferrant:


sbtc-core/src/signer/blockchain/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+        Self {
+            bitcoin_node_rpc_url,
+            stacks_node_rpc_url,
+        }
+    }
+
+    /// Retrieve the current public keys for the signers and their vote ids from the stacks node
+    pub fn public_keys(&self) -> SBTCResult<PublicKeys> {
+        todo!()
+    }
```
netrome, line 30: Nit: More precise naming. Also, I would like to preserve the namespacing here since it contains interesting information, so instead of importing `signer::PublicKeys` we'd just use it with the namespace directly. I'm fine with the import though.

```suggestion
    pub fn signer_public_keys(&self) -> SBTCResult<PublicKeys> {
        todo!()
    }
```

netrome: Something we should add here is reward cycle information. Either on the query or the response, but preferably on the query. I.e.

```
pub fun signer_public_keys(&self, reward_cycle) -> ...
```

That would make the function call idempotent, which would help us build a much more reliable signer.

netrome: Or possibly not completely idempotent, since the broker might not have the information. So we'd return an Option or a Result. But the returned value if any would always be consistent.

sbtc-core/src/signer/blockchain/mod.rs
```diff
@@ -0,0 +1,46 @@
...
+
+    /// Retrieve withdrawal transactions from the stacks node
+    pub fn withdrawal_transactions(&self) -> SBTCResult<Vec<StacksTransaction>> {
+        todo!()
+    }
```
netrome, line 35: Would this be the entire set of withdrawal transactions? I guess we'd like to input some index/cursor to "paginate" the results, or allow the caller to specify which ones to query.

sbtc-core/src/signer/coordinator/fire.rs
```diff
@@ -1,2 +1,28 @@
...
+/// FIRE coordinator
+
+#[derive(Default)]
+pub struct Coordinator {}
+
+impl Coordinate for Coordinator {
+    /// Generate the sBTC wallet public key
+    fn generate_sbtc_wallet_public_key(&self, _public_keys: &PublicKeys) -> SBTCResult<PublicKey> {
+        todo!()
+    }
```
netrome, line 19: Since we know this will abstract away a lot of IO and message passing, we should either split this up into separate functions like "initiate_dkg" and "get_sbtc_wallet_public_key". Optionally I'd be fine with making this async directly.

sbtc-core/src/signer/coordinator/fire.rs
```diff
@@ -1,2 +1,28 @@
...
+    /// Run the signing round for the transaction
+    fn run_signing_round(
+        &self,
+        _public_keys: &PublicKeys,
+        _tx: &BitcoinTransaction,
+    ) -> SBTCResult<(Signature, SchnorrProof)> {
```
netrome, line 25: Same here.

netrome (changes requested):
I'd like to see some changes to the broker interface, and also make sure the functions abstracting away IO takes that into consideration.

sbtc-core/src/signer/blockchain/mod.rs
jferrant: This would be pending withdrawal_transactions because of the get-pending-wallet-peg-outs function. All unprocessed withdrawals from the smart contract.

sbtc-core/src/signer/coordinator/fire.rs
jferrant: generate_sbtc_wallet_public_key is abstracted away. If youw anted the address in the smart contract, you would go to the broker. As for communication via stacker db, the coordinator would need to read internal store that the broker would be in charge of updating I imagine.

stjepangolemac (dismissed):
The thing that bugs me a bit is the introduction of an async trait, especially if we're still in favor of building the signer in an event driven fashion.

If the use case is to wait for the signing round to be done it's very easy to spin up a throwaway thread and communicate the result through an event (channel) back to the main one. Still, it might be a question of singing multiple things at the same time, for which async would probably be the best option.

Anyways, I still want to approve this one, just trying to catch up to the latest consensus about the architecture.

sbtc-core/src/signer/blockchain/mod.rs
```diff
@@ -0,0 +1,324 @@
...
+    /// Bad state
+    BadPegState,
+}
+
+/// Placeholder for important data for a speific signer in a specific cycle
+struct SignerData {
+    /// The amount stacked in the cycle
+    amount: u64,
+    /// the locked stacks balance
+    vote: Option<bool>,
```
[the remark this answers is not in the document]
jferrant: Woops. yep that is a copy paste doc.

sbtc-core/src/signer/coordinator/fire.rs
```diff
@@ -1,2 +1,32 @@
...
+    SBTCResult,
+};
+use bitcoin::{PublicKey, Transaction as BitcoinTransaction};
+use wsts::{bip340::SchnorrProof, common::Signature};
+/// FIRE coordinator
+
+#[derive(Default)]
+pub struct Coordinator {}
+
+#[async_trait]
```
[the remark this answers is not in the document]
jferrant: This was done as an async function as the dkg round for example can take 2 hrs and we should be able to continue processing any backlogged transactions while that operation is going on. However, this PR prob should get closed given the new shift to nakamato focus.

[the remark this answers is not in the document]
jferrant: Or we could merge it and rip out code later as we develop a better understanding of what will be required for the block producer signer.

jferrant:
We are moving towards nakamato instead of mini. Closing this for now as we stub out a new architecture for the block producer signer.

PR #51, opened 2023-08-12
Signer SDK: Add config file parsing

jferrant:


sbtc-core/src/signer/config.rs
```diff
@@ -1,16 +1,245 @@
...
+            auto_deny_addresses_btc: vec![],
+            auto_deny_addresses_stx: vec![],
+            auto_deny_deadline_blocks: 10,
+        }
+    }
+}
+
+impl Config {
+    /// Load the configuration from a toml file
+    pub fn from_path(path: impl AsRef<std::path::Path>) -> SBTCResult<Self> {
```
netrome, line 35: This should probably have it's own result type. I don't think we should abuse the `SBTCResult` type and have it everywhere. This probably just is a IO result, or we can create a specialized custom result type. If you don't want to bother, I suggest defaulting to `anyhow` instead of `SBTCResult`.

sbtc-core/src/signer/config.rs
```diff
@@ -1,16 +1,245 @@
...
+        for address in config.auto_deny_addresses_btc.unwrap_or_default() {
+            auto_deny_addresses_btc.push(BitcoinAddress::from_str(&address).map_err(|e| {
+                SBTCError::InvalidConfig(format!("Failed to parse bitcoin address: {}", e))
+            })?);
+        }
+        for address in config.auto_deny_addresses_stx.unwrap_or_default() {
+            auto_deny_addresses_stx.push(StacksAddress::try_from(address.as_str()).map_err(
+                |e| SBTCError::InvalidConfig(format!("Failed to parse stacks address: {}", e)),
+            )?);
+        }
```
netrome, line 56: Nit: In general I prefer to avoid mutating vectors as much as possible. Here you seem to just be constructing the vectors from iterators, so I'd instead recommend doing something like:

```
let auto_deny_addresses_btc: Result<Vec<_>, _> = config.auto_deny_addresses_btc.into_iter().flatten().map(String::as_str).map(BitcoinAddress::from_str).collect();
```

sbtc-core/src/signer/config.rs
```diff
@@ -1,16 +1,245 @@
...
+    /// Load the configuration from a toml file
+    pub fn from_path(path: impl AsRef<std::path::Path>) -> SBTCResult<Self> {
+        let config_toml = ConfigTOML::from_path(path)?;
+        Self::try_from(config_toml)
+    }
+}
+
+impl TryFrom<ConfigTOML> for Config {
+    type Error = SBTCError;
+    fn try_from(config: ConfigTOML) -> SBTCResult<Self> {
```
netrome, line 43: Also we should have another error type here

sbtc-core/src/lib.rs
```diff
@@ -37,6 +37,9 @@ pub enum SBTCError {
     /// Stacks error
     #[error("Stacks error: {0}")]
     StacksError(#[from] StacksError),
+    /// Invalid Config File
+    #[error("Invalid config file")]
+    InvalidConfig(String),
```
stjepangolemac, line 42: Similar to @netrome I'm also wondering if signer should have it's own error enum and result type. It feels like signer is different enough and has a domain separate enough, but I'm still not sure. What do you think @jferrant?

sbtc-core/src/signer/config.rs
```diff
@@ -1,16 +1,245 @@
...
+            auto_deny_addresses_btc,
+            auto_deny_addresses_stx,
+            auto_deny_deadline_blocks,
+        })
+    }
+}
+
+#[derive(Clone, Deserialize, Debug)]
+/// TOML Configuration for the signer
+pub struct ConfigTOML {
```
stjepangolemac, line 69: Why do we have both Config and ConfigTOML types? If the conversion is fallible can we cover that with a fallible constructor in a single type?

stjepangolemac (approved):
I might be missing some context from the hackaton so my questions might not make sense. It looks good otherwise!

jferrant:
Closing this since moving away from mini towards nakamato instead.

PR #56, opened 2023-08-18
[stacks-core] Stacks transactions

### LINKED ISSUE #23: Stacks transaction APIs
- StacksTransaction
- other types that go with it
- signing utilities
- serialization and deserialization

stjepangolemac:
- adding a bunch of types related to transactions
- still unsure how to proceed - replicate types or try to simplify


stacks-core/src/transaction/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+        Value,
+        NonfungibleConditionCode,
+    ),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Transaction {
+    pub version: TransactionVersion,
+    pub chain_id: u32,
+    pub auth: TransactionAuth,
```
jferrant, line 139: Wonder if these more complicated types should be pulled into their own module similar to how Payload is done.

AshtonStephens:
Is this PR stale? Can we close it?

friedger:
@stjepangolemac would know best.
Related to #80

stjepangolemac:
It's stale and needs to be updated with latest `main` or started from scratch. We need to be able to handle and create Stacks transactions from the SDK.

PR #78, opened 2023-08-28
[sBTC DR] Asset contract deployment

### LINKED ISSUE #73: [sBTC DR] Asset contract
The new sBTC DR will need a simplified asset contract.

#### Definition of done

We have an asset contract defined supporting the SIP-010 FT standard, where:

- [ ] The contract deployer is the only one authorized to mint and burn the token.
- [ ] The mint and burn calls takes a reference to a bitcoin transaction and only materialize if that transaction is mined.

stjepangolemac:
Closes #73

Note, we've left the "only materialize mints/burns if the corresponding bitcoin tx is mined" feature in #73 unimplemented for now. We just maintain the right interface, and will leave that for subsequent clarity work in https://github.com/stacks-network/sbtc/issues/79


netrome (approved):

romeo/asset-contract/contracts/asset.clar
```diff
@@ -0,0 +1,94 @@
...
+    (begin
+        (asserts! (is-contract-owner) (err err-invalid-caller))
+        (ok (var-set bitcoin-wallet-public-key (some public-key)))
+    )
+)
+
+(define-public (mint! (amount uint) (dst principal) (deposit-txid (string-ascii 72)))
+    (begin
+        (asserts! (is-contract-owner) (err err-invalid-caller))
+        ;; TODO #79: Assert deposit-txid exists on chain
```
jferrant, line 36: I would ping Setzeus about this cause am pretty sure this logic already exists in the mini contract.

netrome: True, although Mini does complete validation of the bitcoin transaction and all we need here is to assert that it exists.

Involving Setzeus here makes sense either way. Just noticed I can't ping him. I'll request to add him to the repo.

romeo/src/stacks_client.rs
```diff
@@ -0,0 +1,247 @@
...
+    address::{
+        AddressHashMode, C32_ADDRESS_VERSION_MAINNET_SINGLESIG,
+        C32_ADDRESS_VERSION_TESTNET_SINGLESIG,
+    },
+    types::chainstate::{StacksAddress, StacksPrivateKey, StacksPublicKey},
+};
+
+/// Wrapped Stacks Client which can be shared safely between threads.
+#[derive(Clone, Debug)]
+pub struct LockedClient(Arc<Mutex<StacksClient>>);
```
jferrant, line 31: Is this something that should maybe go in the SDK? I wrote the same thing as part of block producer code (minus the wrapping for mulithreading). Though the signing portion would have to be a trait that could be overwritten to accommodate FROST for example.

stjepangolemac: Yeah this belongs in the SDK but to save a bit of time we've just put it here for now.

jferrant: https://github.com/stacks-network/sbtc/issues/80

jferrant (approved):

PR #83, opened 2023-08-29
[sBTC DR] Stacks client minting and transaction state updates

stjepangolemac:
- minting
- stacks tx status update handling

### DIFF
--- a/romeo/src/system.rs
+++ b/romeo/src/system.rs
@@ -163,7 +163,7 @@ async fn mint_asset(config: &Config, client: LockedClient, deposit_info: Deposit
 
     let recipient = PrincipalData::Standard(StandardPrincipalData(
         deposit_info.recipient.version,
-        deposit_info.recipient.bytes.as_bytes().clone(),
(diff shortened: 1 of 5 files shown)


netrome (approved):

romeo/src/system.rs
```diff
@@ -152,16 +156,99 @@ async fn deploy_asset_contract(config: &Config, client: LockedClient) -> Event {
...
+    let tx_auth = TransactionAuth::Standard(
+        TransactionSpendingCondition::new_singlesig_p2pkh(config.stacks_public_key()).unwrap(),
+    );
+
+    let recipient = PrincipalData::Standard(StandardPrincipalData(
+        deposit_info.recipient.version,
+        *deposit_info.recipient.bytes.as_bytes(),
+    ));
+
+    let function_args = vec![
```
jferrant, line 169: This code will most likely need to be in a function. But assume we can just leave this for when you add the burn logic.

jferrant (approved):

PR #87, opened 2023-08-29
[sBTC DR] Port sbtc-cli and replace blockstack_lib usage with stacks-core

stjepangolemac:
- broadcast
- deposits
- withdrawals
- credential generation
- updated deposit wire format

### DIFF
--- a/sbtc-cli/src/commands/mod.rs
+++ b/sbtc-cli/src/commands/mod.rs
@@ -1,0 +1,5 @@
+pub mod broadcast;
+pub mod deposit;
+pub mod generate;
+pub mod utils;
+pub mod withdraw;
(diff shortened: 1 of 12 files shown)


sbtc-cli/src/commands/mod.rs
```diff
@@ -0,0 +1,5 @@
+pub mod broadcast;
```
jferrant, line 1: Hate to do this...Mind adding forbid missing docs XD

jferrant (approved):
Will reapprove after the forbid missing docs is added :D

sbtc-cli/src/commands/mod.rs
netrome: We should implement a Jacinta-bot to post these comments :joy:

netrome (approved):

PR #92, opened 2023-08-30
[sBTC DR] Deposit flow

stjepangolemac:
- bitcoin block parsing connected to the events and tasks
- contract deployment working
- minting working


sergey-shandar (dismissed):

romeo/src/stacks_client.rs
```diff
@@ -163,7 +163,7 @@ impl StacksClient {
             .json()
             .await?;
 
-        Ok(fee_rate * tx_len)
+        Ok(fee_rate * tx_len * 100)
```
AshtonStephens, line 166: Where does this 100 come from and can we store it in a constant so that is has a meaningful variable name?

romeo/src/state.rs
```diff
@@ -105,7 +131,7 @@ where
 pub fn bootstrap(state: &State) -> Task {
     match state.contract {
         None => Task::CreateAssetContract,
-        Some(_) => Task::FetchBitcoinBlock(state.block_height),
+        Some(_) => Task::FetchBitcoinBlock(state.block_height.map(|block_height| block_height + 1)),
```
jferrant, line 134: Only because you do this in two different spots (line 134 and line 174), it might be worthwhile making this a helper function like "next_block_height" to prevent a typo down the line.

romeo/src/stacks_client.rs
```diff
@@ -180,15 +180,30 @@ impl StacksClient {
...
         let mut rng = thread_rng();
         let random_string: String = (0..16).map(|_| rng.sample(Alphanumeric) as char).collect();
 
-        // We need to make sure node returns the uncached nonce, so we add a cachebuster
+        let mut query = url
+            .query()
+            .map(|query| query.to_string())
+            .unwrap_or_default();
+
+        query = if query.is_empty() {
```
AshtonStephens, line 192: This is neurotic, but thoughts on something like:

```rust
query = match query.is_empty() {
  true => format!("cachebuster={}", random_string),
  false => format!("{}&cachebuster={}", query, random_string)
}
```

I prefer match when the conditional returns a value.

romeo/src/stacks_client.rs
```diff
@@ -180,15 +180,30 @@ impl StacksClient {
...
+
+        url.set_query(Some(&query));
+
+        url
+    }
+
+    fn nonce_url(&self) -> reqwest::Url {
         let path = format!(
-            "/extended/v1/address/{}/nonces?cachebuster={}",
+            "/extended/v1/address/{}/nonces",
```
AshtonStephens, line 205: Thoughts on putting this into a constant?

romeo/src/state.rs
```diff
@@ -37,18 +42,31 @@ pub struct DepositInfo {
...
 impl Deposit {
-    fn mint(&mut self, _block_height: u64) -> Option<Task> {
-        todo!();
+    fn mint(&self) -> Option<Task> {
+        match self.mint.as_ref() {
+            Some(Response {
+                status: TransactionStatus::Broadcasted,
+                txid,
+            }) => Some(Task::CheckStacksTransactionStatus(*txid)),
+            // TODO: Think about removing deposits at this stage
```
AshtonStephens, line 58: IMO TODOs should always have a corresponding ticket so that if we decide never to do it or do decide to do it the effort is tracked. What are your thoughts?

romeo/src/state.rs
```diff
@@ -125,50 +151,105 @@ pub fn update(config: &Config, state: State, event: Event) -> (State, Vec<Task>)
...
-    let deposits = parse_deposits(&block);
+fn process_bitcoin_block(config: &Config, mut state: State, block: Block) -> (State, Vec<Task>) {
+    let deposits = parse_deposits(config, &block);
     let withdrawals = parse_withdrawals(&block);
 
     state.deposits.extend_from_slice(&deposits);
     state.withdrawals.extend_from_slice(&withdrawals);
 
-    state.block_height = block
+    let new_block_height = block
```
AshtonStephens, line 166: Does it also work to do this?

```
let new_block_height: u32 = block
        .bip34_block_height()
        .expect("Failed to get block height");
```

If this is valid it would be better to explicitly type the variable as opposed to typing the output and implicitly typing the variable.

romeo/src/state.rs
netrome: Agreed. I've tried to enforce it throughout the code base but occasionally something slips through.

romeo/src/stacks_client.rs
stjepangolemac: It's still unclear what is the right number for this. We've observed some transactions being stuck in the mempool for days and have suspected it might have something to do with the fees. On the other hand testnet nodes all return `1` as the fee rate constantly. At the moment this is just a random multiplier but as we find out more about testnet fees we can make sure to tidy it up.

I've created this issue https://github.com/stacks-network/sbtc/issues/98.

romeo/src/state.rs
stjepangolemac: Yep can do that 👍

romeo/src/stacks_client.rs
stjepangolemac: It's not that I don't like it, but considering we're matching on a boolean it doesn't feel idiomatic. Anyways, I'll try it and see if cargofmt keeps it formatted as your suggestion. I do dislike having a lot of open / closing bracket lines.

romeo/src/stacks_client.rs
stjepangolemac: I usually put stuff in constants when they're used more than once. I can't imagine this being used anywhere else.

romeo/src/state.rs
stjepangolemac: Created a separate issue https://github.com/stacks-network/sbtc/issues/99.

romeo/src/state.rs
stjepangolemac: I don't think that will work. The cast needs to be explicit.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3241a4156cd68f163d6053b829a9bb7f

romeo/src/stacks_client.rs
stjepangolemac: Yep I did what you suggested. It looks nicer.

romeo/src/state.rs
stjepangolemac: And added the issue number to the TODO comment.

stjepangolemac:
I'll need you to review again when you find time 🙏

netrome (approved):

friedger (approved):
LGTM!

radicleart (commented):
I'm not a rust developer but this looks good to me.

I've created an issue to update the bridge code to align with the deposit flow described here - see https://github.com/Stacks-Builders/sbtc-bridge-api/issues/67

PR #107, opened 2023-09-04
[sBTC DR] Withdrawal flow

stjepangolemac:
- Parses withdrawal requests from the bitcoin state
- Burns sBTC
- Withdraws BTC to the destination address
- It now separately tracks Bitcoin and Stacks blocks
- Requires RPC Bitcoin node access
- Requires Electrum service access


romeo/asset-contract/contracts/clarity-bitcoin-mini.clar
radicleart, line 1: There's an issue with this change - verify-block-header should be called and DEBUG_MODE needs to be set to false on testnet deployment.

radicleart (commented):
on deploying the contracts to testnet it came out that DEBU_MODE In calrity-bitcoin-mini needs to be switched off and `was-tx-mined` needs to call `verify-block-header`

stacks-core/src/address.rs
```diff
@@ -35,6 +32,40 @@ pub enum AddressVersion {
...
+/// Supported stacks address versions
+#[repr(u8)]
+#[derive(FromRepr, EnumIter, EnumString, PartialEq, Eq, Copy, Clone, Debug)]
+#[strum(ascii_case_insensitive)]
+pub enum AddressKind {
+    /// Pay to public key hash
+    P2PKH = 0,
+    /// Pay to script hash
+    P2SH = 1,
+    /// Pay to witness public key hash
```
friedger, line 44: Maybe add a comment that this is the only AddressKind used in Stacks for single accounts

friedger (changes requested):
Looks good apart from the change to clarity-bitcoin-mini

romeo/asset-contract/contracts/clarity-bitcoin-mini.clar
stjepangolemac: Okay I guess it makes sense for debug mode to be disabled by default.

stacks-core/src/address.rs
stjepangolemac: It's actually a p2pkh if you look at the `Address::from_public_key` function. I'm not sure at the moment which one should it be.

AshtonStephens:
For my own curiosity - is there a diagram of this flow anywhere?

stacks-core/src/address.rs
friedger: p2pkh looks correct

stjepangolemac:
> For my own curiosity - is there a diagram of this flow anywhere?

No, is that something we'd like to have in the docs or only here to make the review clearer?

stacks-core/src/address.rs
stjepangolemac: I removed AddressKind completely. It's probably a remnant of the code I used to signal what kind of Stacks address should be used to burn from. We now by convention always use P2PKH for that.

.gitignore
```diff
@@ -4,4 +4,4 @@ coverage
 romeo/testing/
 romeo/asset-contract/.test
 romeo/asset-contract/.coverage
-**/.DS_Store
+**/romeo.log
```
friedger, line 7: DS_Store should stay

romeo/asset-contract/contracts/clarity-bitcoin-mini.clar
```diff
@@ -2,7 +2,7 @@
 
 ;; it can only check if a txid is part of a burn chain block
 
-(define-constant DEBUG-MODE true)
+(define-constant DEBUG-MODE false)
```
friedger, line 5: Now unit tests fail

friedger (changes requested):

romeo/asset-contract/contracts/clarity-bitcoin-mini.clar
stjepangolemac: I'll revert the debug mode change as that's being fixed in a separate PR.

friedger (approved):
All good! Great work!

AshtonStephens:
Lots of code cov warnings here - is none of this actually covered? Also - how are we sure that this works? What local testing did you run so I can recreate it?

devenv/utils/deposit.sh
```diff
@@ -0,0 +1,21 @@
...
+
+json=$(sbtc deposit \
+    -w $btc_wif \
+    -n regtest \
+    -r $stacks_address \
+    -a $amount \
+    -d $btc_p2tr_address \
+    -u localhost:60401)
+
+tx=$(echo -n $json | jq -r .hex)
```
friedger, line 19: Can we show some output if the deposit fails?

devenv/utils/mine_btc.sh
```diff
@@ -0,0 +1,13 @@
...
+
+num_blocks=$1
+dir="$(dirname "$0")"
+
+if [[ -z "$num_blocks" ]]; then
+    echo "Usage: ./script.sh [num_blocks]"
+    exit 1
+fi
+
+btc_address=$(source $dir/get_credentials.sh | jq -r '.credentials["0"].bitcoin.p2wpkh.address')
```
friedger, line 11: We need also bitcoins for p2pkh:

btc_address=$(source $dir/get_credentials.sh | jq -r '.credentials["0"].bitcoin.p2pkh.address')

AshtonStephens:
@stjepangolemac since you're out next week it may make the most sense to merge this now and then iterate on any bugs that we catch with patches to the main branch.

AshtonStephens (approved):
Approving so this is merged knowing it lacks coverage and isn't perfect.

romeo/src/bitcoin_client.rs
```diff
@@ -0,0 +1,218 @@
...
+			return Err(anyhow::anyhow!("Password is empty"));
+		}
+
+		url.set_username("").unwrap();
+		url.set_password(None).unwrap();
+
+		let client =
+			RPCClient::new(url.as_ref(), Auth::UserPass(username, password))?;
+
+		Ok(spawn_blocking(move || f(client)).await?)
```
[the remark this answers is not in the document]
stjepangolemac: The underlying client implementation is blocking and by spawning this to a separate blocking worker thread we're not blocking the runtime.

devenv/utils/deposit.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
stjepangolemac: Yep, sure.

devenv/utils/deposit.sh
stjepangolemac: Yeah can try to do something like that.

stjepangolemac: Added this to https://github.com/stacks-network/sbtc/pull/175

devenv/utils/mine_btc.sh
stjepangolemac: Added this to https://github.com/stacks-network/sbtc/pull/175

devenv/utils/deposit.sh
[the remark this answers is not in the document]
stjepangolemac: Added this to https://github.com/stacks-network/sbtc/pull/175

romeo/src/bitcoin_client.rs
```diff
@@ -0,0 +1,218 @@
...
+#[derive(Debug, Clone)]
+pub struct Client(Config);
+
+impl Client {
+	/// Create a new RPC client
+	pub fn new(config: Config) -> anyhow::Result<Self> {
+		Ok(Self(config))
+	}
+
+	async fn execute<F, T>(
```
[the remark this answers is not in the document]
stjepangolemac: I can add some docs but will add them in https://github.com/stacks-network/sbtc/pull/175

Issue #79, opened 2023-08-28
[sBTC DR] Assert deposit/witdhdrawal tx exists in mint/burn contract calls


netrome:
### Definition of done
The sBTC DR asset contract does not allow mints and burns that do not refer to existing bitcoin transactions

MarvinJanssen:
How was the interface decided? #78 mentioned "[...] We just maintain the right interface [...]". Is this "right interface" intended to only be for Romeo or beyond? I understood that for Romeo we want to define an interface that is stable.

If that is the case, we need to work closely on defining that interface.

I can take on this task but I need to know the exact requirements. We have all the logic to check if a Bitcoin transaction is mined. It just needs to be moved over and adapted.

netrome:
> How was the interface decided? #78 mentioned "[...] We just maintain the right interface [...]". Is this "right interface" intended to only be for Romeo or beyond? I understood that for Romeo we want to define an interface that is stable.
> 
> If that is the case, we need to work closely on defining that interface.
> 
> I can take on this task but I need to know the exact requirements. We have all the logic to check if a Bitcoin transaction is mined. It just needs to be moved over and adapted.

Good questions @MarvinJanssen . We kept the same interface to the `mint` and `burn` calls from the previous alpha work, where both of them takes a transaction ID and a principal as input. If you anticipate that we need other things in the interface, this is a good time to update it or redefine it, but to the best of my knowledge that is the only information the mint and burn calls will need after the consensus breaking release as well.

Two additional points to keep in mind as well:
1. The interface is in no way frozen and we will update it as we discover new things, as long as we maintain parity between Nakamoto sBTC and DR sBTC.
2. The `mint!` and `burn!` function signatures are not user-facing which gives us more flexibility to update them in the future. However, users/developers may rely on events emitted from them so we should be more careful about those.

friedger:
To verify that a tx exists the contract needs a block height, raw tx,... up to 10 parameters as described in https://clarity-bitcoin.gitbook.io/clarity-bitcoin-library/clarity-functions

Can we expect a certain format of the btc transactions? @netrome 
Are they always segwit txs? Always N inputs and M outputs?

netrome:
> To verify that a tx exists the contract needs a block height, raw tx,... up to 10 parameters as described in https://clarity-bitcoin.gitbook.io/clarity-bitcoin-library/clarity-functions
> 
[...]

PR #108, opened 2023-09-05
feat: verify if passed txid is mined for mint! and burn!

MarvinJanssen:
Addresses #79, but the exact purpose of the requirement remains unclear. Since only the contract owner can `mint!` and `burn!`, and these functions will not be part of the API, why is the check truly needed here?

Think this is enough @friedger @FriendsFerdinand? I'll add tests if you agree with the approach.


friedger:
Looks good, this approach. The verification is needed so that we know that sbtc is minted on the stacks chain using the same bitcoin fork as the bitcoin node.

romeo/asset-contract/contracts/asset.clar
```diff
@@ -5,16 +5,16 @@
 
 ;; traits
 ;;
-(impl-trait 'ST1NXBK3K5YYMD6FD41MVNP3JS1GABZ8TRVX023PT.sip-010-trait-ft-standard.sip-010-trait)
+(impl-trait 'SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE.sip-010-trait-ft-standard.sip-010-trait)
```
jo-tm, line 8: Do these traits are only in Mainnet or can be deployed in Devnet with the same principal address?

jo-tm (approved):
approved with a minor question on hardoced principals...

romeo/asset-contract/contracts/asset.clar
friedger: They can be removed alltogether

friedger:
@MarvinJanssen Can you use branch 79-proof as your base? It contains the changes to call your function.

FriendsFerdinand (approved):

friedger (approved):
LGTM

Tests still open.

romeo/asset-contract/contracts/asset.clar
```diff
@@ -90,5 +118,9 @@
 ;; private functions
 ;;
 (define-private (is-contract-owner)
-    (is-eq (var-get contract-owner) tx-sender)
+    (ok (asserts! (is-eq (var-get contract-owner) contract-caller) err-invalid-caller))
```
friedger, line 121: Is this more efficient?

For a return ok, I would expect the name of the function something like `check-is-contract-owner`.

MarvinJanssen: Not sure if it is more efficient but it definitely makes using the function easier to use (with a `try!`)—and there is no chance you return the wrong error value somewhere. It is a standard pattern I follow.

MarvinJanssen:
I'll do tests in the next PR.

MarvinJanssen:
@friedger I didn't notice you changed the base so now I merged it into `79-proof` instead of `master`. I'm not sure how to undo it on Github.

PR #174, opened 2023-09-27
[sBTC DR] filter by sbtc wallet address

### LINKED ISSUE #167: [sBTC DR] verify sbtc wallet address in btc txs
Only handle btc deposit and withdrawal txs for the current sbtc wallet address.

friedger:
## Description

This PR
* filters deposit and withdrawal requests by sbtc wallet address
* fixes #167


romeo/src/state.rs
```diff
@@ -232,10 +238,23 @@ fn process_bitcoin_block(
 	bitcoin_height: u32,
 	block: Block,
 ) -> (State, Vec<Task>) {
```
CAGS295, line 240: Question: why does block processing depend on the config? Shouldn't the peg wallet be derived from something on-chain? I find odd filtering deposits and withdrawals based on an immutable config field.

romeo/src/state.rs
```diff
@@ -232,10 +238,23 @@ fn process_bitcoin_block(
...
-	state
-		.deposits
-		.extend(parse_deposits(config, bitcoin_height, &block));
-	state.withdrawals.extend(parse_withdrawals(config, &block));
+	state.deposits.extend(
+		parse_deposits(config, bitcoin_height, &block)
+			.into_iter()
+			.filter(|deposit| {
+				deposit.info.sbtc_wallet_address == config.sbtc_wallet_address()
```
CAGS295, line 245: Consider storing `config.sbtc_wallet_address()` before iterating.

romeo/src/state.rs
friedger: The romeo engine is the single source of truth and the on chain info is only a copy of it for information.

Romeo has no (and will not have) function to change the sbtc wallet address while running.

romeo/src/state.rs
friedger: Addressed in 41854d9f6ae44a968b7fdb24184dd85041492503

CAGS295 (approved):

friedger:
replaced by #185

PR #175, opened 2023-09-27
Romeo should verify and update the public key on startup

stjepangolemac:
- [x] Will set the public key when contract is discovered
- [x] State handling refactored to make it clearer and more understandable
- [x] Logging cleaned up and now uses sane defaults (no need to set RUST_LOG to see something)


romeo/src/state.rs
```diff
@@ -12,332 +12,637 @@ use sbtc_core::operations::{
...
-#[derive(Debug, Default, serde::Serialize, serde::Deserialize)]
-pub struct State {
-	deposits: Vec<Deposit>,
-	withdrawals: Vec<Withdrawal>,
-	stacks_block_height: Option<u32>,
-	bitcoin_block_height: Option<u32>,
-}
+/// Romeo internal state
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+pub enum State2 {
```
friedger, line 25: Why `State2` and not `State` ?

friedger (dismissed):
Looks good to me.

radicleart (approved):
Looks good to me.

friedger:
Needs rebase :-)

romeo/src/state.rs
stjepangolemac: Leftover from the refactor, will fix.

stjepangolemac:
> Looks good to me.
> 
> Q. Do we need to set the 33 byte public key? Given we always strip the first (x-only) byte to get the schnorr compatible one - is there a case where its needed?

You mean why not write 32 bytes only?

stjepangolemac:
I've updated the branch and could use your reapprovals @radicleart and @friedger.

friedger (approved):
LGTM

radicleart (approved):
lgtm

friedger:
@AshtonStephens Mike's review does not count.

stjepangolemac:
> @AshtonStephens Mike's review does not count.

Does anyone know why is this happening?

AshtonStephens:
> > @AshtonStephens Mike's review does not count.
> 
> Does anyone know why is this happening?

I've added Mike as a maintainer

Issue #182, opened 2023-09-28
[sBTC DR] devenv does not find stacks-api


friedger:
```
023-09-28T12:41:21.104514Z TRACE romeo::stacks_client: Retrying in 77.588151199s after error: error sending request for url (http://stacks-api:3999/extended/v1/contract/ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.asset): error trying to connect: tcp connect error: Connection refused (os error 111)
```

CAGS295:
Closing this as stale. Revive it if it shows up again.

friedger:
I appeared while running dev/integration, but maybe my computer was out of space. I'll do more investigations.

PR #192, opened 2023-09-29
Propagate error when an error field is present in the request response.

CAGS295:
## Description

When the request fails, I expect the error field to be present and propagated as an error.

Expected:
```
2023-09-29T05:40:19.736443664Z thread 'tokio-runtime-worker' panicked at 'Could not get block height: cannot find contract by ID ST2GEN9CVCJNZPDH7Z7NPWRG6Z67GH3WWYAS4DGRA.sbtc-alpha-romeo-testing', romeo/src/system.rs:192:10
```

### DIFF
--- a/romeo/src/stacks_client.rs
+++ b/romeo/src/stacks_client.rs
@@ -2,7 +2,7 @@
 
 use std::{io::Cursor, sync::Arc, time::Duration};
 
-use anyhow::anyhow;
+use anyhow::{anyhow, Error};
 use blockstack_lib::{
 	burnchains::Txid as StacksTxId,
 	chainstate::stacks::{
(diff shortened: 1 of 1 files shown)


friedger (approved):
The same can be applied to `get_bitcoin_block_height` and other requests

CAGS295:
Yes. I'm trying to figure out how to proceed. I wanted to do that but felt the whole file could use refactoring. Then, I wanted to add tests. But touching many methods means fixing test coverage for everything. It could quickly get out of scope. I'm trying to get devenv to work stock. I want to have parity with your setup until #182 shows. That's why I'm trying to fix just the minimum.

PR #173, opened 2023-09-27
[sBTC DR] Add check for contract versions via clarinet-sdk

friedger:
## Description
This PR 
* adds a unit test to check that the contract versions of `clarity-bitcoin-mini` differ only in DEBUG flag
* introduces clarinet-sdk to this repo
* fixes #172

PR #206, opened 2023-09-30
Only use btc tx once

### LINKED ISSUE #116: [sBTC DR] binary can use the same tx several times
We have a clarity tests that shows that the binary can use the same bitcoin tx serveral times to mint or burn. Can we trust the binary that it does not rerun the mint or burn flow several times? E.g. when restarting?

image

This is maybe a discussion: Should we add a registry of txs used for mint and burn?

friedger:
## Summary of Changes
This PR 
* changes the asset contract so that mints and burns with the same btc tx will fail when used more than once
* fixes #116 

### Risks
Engine might not handle failed mints and burns correctly.

Users might think that the contract verifies all of the content of the text. It does not, the contract only verifies that the btc tx was not yet used for minting or burning.

manual in devenv

Automated integration test

## Checklist:

### DIFF
--- a/romeo/asset-contract/contracts/asset.clar
+++ b/romeo/asset-contract/contracts/asset.clar
@@ -17,11 +17,15 @@
 ;;
 (define-constant err-invalid-caller (err u4))
 (define-constant err-forbidden (err u403))
+(define-constant err-btc-tx-already-used (err u500))
 
 ;; data vars
 ;;
 (define-data-var contract-owner principal tx-sender)
 (define-data-var bitcoin-wallet-public-key (optional (buff 33)) none)
+
+;; stores all btc txids that have been used to mint or burn sBTC
+(define-map amounts-by-btc-tx (buff 32) int)
 
 ;; public functions
 ;;
@@ -54,6 +58,7 @@
     (begin
         (try! (is-contract-owner))
         (try! (verify-txid-exists-on-burn-chain deposit-txid burn-chain-height merkle-proof tx-index tree-depth block-header))
+        (asserts! (map-insert amounts-by-btc-tx deposit-txid (to-int amount)) err-btc-tx-already-used)
         (try! (ft-mint? sbtc amount destination))
         (print {notification: "mint", payload: deposit-txid})
         (ok true)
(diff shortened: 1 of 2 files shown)


setzeus (approved):
Pulled & ran locally, can confirm all unit tests passing - LGTM.

romeo/asset-contract/contracts/asset.clar
```diff
@@ -72,6 +77,7 @@
     (begin
         (try! (is-contract-owner))
         (try! (verify-txid-exists-on-burn-chain withdraw-txid burn-chain-height merkle-proof tx-index tree-depth block-header))
+        (asserts! (map-insert amounts-by-btc-tx withdraw-txid (* -1 (to-int amount))) err-btc-tx-already-used)
```
CAGS295, line 80: I've seen lately that folks are replacing asserts! with assert-eq. Not sure if this is one of those cases.

romeo/asset-contract/tests/asset_test.clar
```diff
@@ -9,6 +9,8 @@
 
 (define-constant err-invalid-caller (err u4))
 (define-constant err-forbidden (err u403))
+(define-constant err-btc-tx-already-used (err u500))
+(define-constant err-expected-err (err u9999))
```
CAGS295, line 13: I don't see this error used anywhere.

romeo/asset-contract/contracts/asset.clar
setzeus: We're doing this (using our custom (assert-eq ...) strictly in the *test* contracts only, not actual Clarity contracts.

setzeus (changes requested):
Additional comment but re-reading everything here I'm not sure why we now have two clarity-bitcoin-mini files? 

We now have our original one & a "boot" version with v little difference. I don't think that was part of this commit history but wanted to flag that we likely have a duplicate file - is this intentional @friedger ?

friedger:
@setzeus we have two versions of the contract, one for unit tests and one for deployment.

Related issue #173

setzeus (approved):

romeo/asset-contract/tests/asset_test.clar
friedger: Has been removed now.

setzeus (approved):
Pulled & working - LGTM!

MarvinJanssen (approved):

PR #229, opened 2023-10-05
utils: use container based cli

EmbeddedAndroid:
## Summary of Changes

The util scripts currently are using host binaries for running commands against devnet. This can be confusing to users as we also provide wrappers for accessing the same binaries running inside the container devenv.

Some examples are:

https://github.com/stacks-network/sbtc/blob/main/devenv/bitcoin/bin/bitcoin-cli

https://github.com/stacks-network/sbtc/blob/main/devenv/sbtc/bin/sbtc

Implements: https://github.com/stacks-network/sbtc/issues/224

This PR switches the utils scripts to use the wrappers for container binaries. For the binaries running inside devenv, it must use the service hostnames otherwise localhost will cause them to fail.

### Risks

The biggest risk I see is, if the sbtc container (romeo) crashes, the sbtc command: https://github.com/stacks-network/sbtc/blob/main/devenv/sbtc/bin/sbtc will not be able to run, as the container stopped.

### How were these changes tested?

Tested locally with devenv

### What future testing should occur?

We need to automate the entire flow via https://github.com/stacks-network/sbtc/pull/215

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #237, opened 2023-10-05
miner updates

EmbeddedAndroid:
## Summary of Changes

As discussed, we want to make it easier to fund the stacks deployer wallet.

Premine both the stacks miner wallet and stacks deployer wallets, and continue to mine to the deplorer's wallet.

This PR depends on https://github.com/stacks-network/sbtc/pull/229

### Risks

The risk is low, as this was tested end to end with devenv

```
./up.sh
**wait for stacks to start mining**
./utils/deploy_contracts.sh
docker compose down sbtc
docker compose up sbtc -d
./utils/deposit.sh
./utils/withdrawal.sh
```

Automation of end to end deposit and withdraw flows.

## Checklist:

### DIFF
--- a/devenv/miner/docker/entrypoint.sh
+++ b/devenv/miner/docker/entrypoint.sh
@@ -12,13 +12,16 @@ curl --user devnet:devnet --data-binary '{"jsonrpc": "1.0", "id": "curltest", "m
 echo "devnet bitcoin wallet created"
 # Import miner address
 curl --user devnet:devnet --data-binary '{"jsonrpc": "1.0", "id": "curltest", "method": "importaddress", "params": ["n3GRiDLKWuKLCw1DZmV75W1mE35qmW2tQm"]}' -H 'content-type: text/plain;' http://bitcoin:18443/
-# Mine the first N blocks
+# Mine the first N blocks to the stacks miner wallet address
 curl --user devnet:devnet --data-binary '{"jsonrpc": "1.0", "id": "curltest", "method": "generatetoaddress", "params": ['''${INIT_BTC_BLOCKS}''', "n3GRiDLKWuKLCw1DZmV75W1mE35qmW2tQm"]}' -H 'content-type: text/plain;' http://bitcoin:18443/
-echo "mined initial blocks"
-# Mine a single block every 10 seconds
+echo "mined initial blocks to stacks miner wallet"
+# Mine the first N blocks to the stacks deployer wallet address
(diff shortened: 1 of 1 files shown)


CAGS295:
Can we reiterate what accounts need funding for regular sBTC operation? Also, @friedger, can you refresh your concern about nonce issues?

Here is what I understood.
1. Mining empty blocks with the developer address will cause nonce issues.
2. The developer address requires initial funds but not necessarily mining continuously.
3. It would be good to stick with the current miner address; I forgot why.

friedger:
My understanding 

**btc**
Deployer does not need btc

Alice needs btc. 

Stbc wallet needs btc for fees if withdrawal request do not include enough fees.

Stacks miner needs btc.

**stx**
Deployer needs stx (funded on startup) 

Alice does not need stx

Sbtc wallet does not need stx, stx txs are send from deployer. 

Stx miner does not need stx.

Re nonce issue, I can't find an argument, why nonce issues might occur. Maybe we just ignore my comment.

CAGS295:
> My understanding
> 
> **btc** Deployer does not need btc
> 
> Alice needs btc.
> 
> Stbc wallet needs BTC for fees if withdrawal requests do not include enough fees.
> 
> Stacks miner needs btc.
> 
> **sex** Deployer needs sex (funded on startup)
> 
> Alice does not need sex
> 
> Sbtc wallet does not need stx; stx txs are sent from deployer.
> 
> Stx miner does not need sex.
> 
> Re nonce issue, I can't find an argument why nonce problems might occur. Maybe we should just ignore my comment.

After digesting your comments, I think it would be best to drop this PR and add an `--address` option to the `mine_btc.sh`. The way utils scripts are used is an integration test flow. So we can go further and hardcode it instead of having `--address.`

devenv/utils/withdraw.sh
```diff
@@ -36,4 +36,4 @@ fi
 
 tx=$(echo -n $json | jq -r .hex)
 
-sbtc broadcast localhost:60401 $tx | jq -r .
```
AshtonStephens, line 39: Why did this work before?

friedger: We all had installed sbtc binary locally..

friedger (approved):
LGTM

The setup about where and when BTCs are mined should be documented in https://github.com/stacks-network/sbtc-docs/blob/master/src/sbtc-releases/sbtc-dev/get-started-on-devnet.md

CAGS295:
@friedger 

> My understanding
> 
> **btc** Deployer does not need btc
> 

Why are we changing the miner to fund the deployer if the deployer does not need btc?

friedger:
Let's move the changes to funding into a new pr

Issue #271, opened 2023-10-12
[sBTC DR] Romeo crash when stacks node not available temporarily


friedger:
### Expected Behavior
Romeo handles failed requests to the stacks node gracefully and tries again

### Actual Behavior
Romeo http request thread crashes and the event loop is broken. After the thread crash only bitcoin blocks are fetched.

Error log:
```
panicked at 'Could not get Stacks block: Could not parse response JSON, URL is https://stacks-node-api.testnet.stacks.co/extended/v1/block/by_height/128953, status is 522
```

### Steps to Reproduce

Launch romeo with your own token on testnet and wait until the hiro node answers with 522.

### Log

``` 
2023-10-12T13:48:20.937332Z  INFO update: romeo::state: Processing event=StacksBlock(128949)
2023-10-12T13:48:20.937401Z  INFO spawn: romeo::system: Spawning task=FetchStacksBlock(128950)
2023-10-12T13:54:36.103059Z  INFO update: romeo::state: Processing event=BitcoinBlock(2532357)
2023-10-12T13:54:36.123679Z  INFO spawn: romeo::system: Spawning task=FetchBitcoinBlock(2532358)
2023-10-12T13:54:43.368060Z  INFO update: romeo::state: Processing event=BitcoinBlock(2532358)
2023-10-12T13:54:43.369051Z  INFO spawn: romeo::system: Spawning task=FetchBitcoinBlock(2532359)
2023-10-12T13:54:53.764960Z  INFO update: romeo::state: Processing event=StacksBlock(128950)
2023-10-12T13:54:53.765031Z  INFO spawn: romeo::system: Spawning task=FetchStacksBlock(128951)
2023-10-12T14:02:47.707777Z  INFO update: romeo::state: Processing event=BitcoinBlock(2532359)
2023-10-12T14:02:47.709436Z  INFO spawn: romeo::system: Spawning task=FetchBitcoinBlock(2532360)
2023-10-12T14:04:13.387532Z  INFO update: romeo::state: Processing event=StacksBlock(128951)
2023-10-12T14:04:13.387817Z  INFO spawn: romeo::system: Spawning task=FetchStacksBlock(128952)
2023-10-12T14:05:28.845569Z  INFO update: romeo::state: Processing event=BitcoinBlock(2532360)
2023-10-12T14:05:28.849823Z  INFO spawn: romeo::system: Spawning task=FetchBitcoinBlock(2532361)
[...]

PR #276, opened 2023-10-13
Add 429 & 522 handling with retry to romeo

AshtonStephens:
### Context

While setting up sBTC for testnet the initialization stage of Romeo makes a ton of block fetch requests in close succession
which will trigger a 429 response unless you have an api key with nearly no throttle. This PR 

### Risks

This will cause status errors that used to be let through to panic - if we had a special way that we handled those cases
we will now be skipping them

No unit tests unfortunately, but there should be. I've run this on testnet and it does what it should.

The caveat is that it will panic when there's a 404 error, which it didn't before, but Romeo keeps going and 
we probably should have been panicking there in the first place.

Same tests as the rest of Romeo

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] My changes generate no new warnings

### DIFF
--- a/romeo/src/bitcoin_client.rs
+++ b/romeo/src/bitcoin_client.rs
@@ -285,7 +285,8 @@ mod tests {
 		// expect sbtc wallet to be p2tr of mnemonic
 		let expected_sbtc_wallet =
 			"tb1pte5zmd7qzj4hdu45lh9mmdm0nwq3z35pwnxmzkwld6y0a8g83nnq6ts2d4";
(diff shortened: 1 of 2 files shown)


stjepangolemac (dismissed):

stjepangolemac:
Can you also include status 522, that would fix this issue too:

https://github.com/stacks-network/sbtc/issues/271

AshtonStephens:
Linter is failing on a comment that isn't in my local source?

stjepangolemac (approved):

CAGS295 (approved):

PR #285, opened 2023-10-13
Add manual workflow to deploy docker images to blockstack

AshtonStephens:
## Summary of Changes

Add manual docker image publishing

### Risks

Few risks. It could provide faulty images and make people believe their code is broken when it's just the images.

Tested on an x86 computer. Has issues with ARM64

Should find a way to generate arm64 images. 

## Checklist:

- [ ] I have made corresponding changes to the documentation

### DIFF
--- a/.github/workflows/publish-to-dockerhub.yml
+++ b/.github/workflows/publish-to-dockerhub.yml
@@ -33,7 +33,7 @@ jobs:
           file: ./devenv/sbtc/docker/Dockerfile
           # TODO: Set up docker publishing to use the tag name when publishing and to do
           # something else when triggered manually. $\{{ github.event.release.tag_name }}
-          tags: blockstack/sbtc:sbtc
+          tags: blockstack/sbtc:devenv-sbtc
           push: true
           platform: amd64
 
@@ -75,5 +75,5 @@ jobs:
           push: true
           # TODO: Set up docker publishing to use the tag name when publishing and to do
           # something else when triggered manually. $\{{ github.event.release.tag_name }}
-          tags: blockstack/sbtc:${{ matrix.service }}
+          tags: blockstack/sbtc:devenv-${{ matrix.service }}
           platform: amd64
--- a/.github/workflows/publish-to-dockerhub.yml
+++ b/.github/workflows/publish-to-dockerhub.yml
@@ -1,0 +1,79 @@
+
+name: Publish Docker image
+
+on:
+  push:
+    branches: build-docker-images
+  release:
+    types: [published]
+  # Allow workflow to be triggered manually.
+  workflow_dispatch:
+
+jobs:
+  push_sbtc_to_registry:
+    name: Push sBTC Docker Image to Docker Hub
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v4
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+      - name: Setup Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v5
+        with:
+          # We use ubuntu-latest which implies amd64
+          context: "{{ defaultContext }}"
+          file: ./devenv/sbtc/docker/Dockerfile
+          # TODO: Set up docker publishing to use the tag name when publishing and to do
+          # something else when triggered manually. $\{{ github.event.release.tag_name }}
+          tags: blockstack/sbtc:sbtc
+          push: true
+          platform: amd64
+
+  push_services_to_registry:
+    strategy:
+      matrix:
+        service: [
+          electrs,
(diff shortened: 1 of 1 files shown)


stjepangolemac:
How would this play out locally if we pull some changes, try spinning up devenv and the images are not built yet in the CI?

AshtonStephens:
It would play out poorly - you would need to run `./build.sh` locally to overwrite the ones here. For most users, this will be fine. 

It might be worth making a separate `devenv` repo that doesn't only pulls in the images. At a point, our code should be stable

.github/workflows/publish-to-dockerhub.yml
```diff
@@ -0,0 +1,79 @@
...
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v5
+        with:
+          # We use ubuntu-latest which implies amd64
+          context: "{{ defaultContext }}"
+          file: ./devenv/sbtc/docker/Dockerfile
+          # TODO: Set up docker publishing to use the tag name when publishing and to do
+          # something else when triggered manually. $\{{ github.event.release.tag_name }}
+          tags: blockstack/sbtc:sbtc
```
CAGS295, line 36: It should be explicit it is Devenv's image.

.github/workflows/publish-to-dockerhub.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Setup Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v5
+        if: ${{ matrix.service }} == sbtc
```
CAGS295, line 69: Is sbtc not in the matrix?

AshtonStephens: sBTC is above, because it's built with the sources here I put it separately.

.github/workflows/publish-to-dockerhub.yml
AshtonStephens: One note here is that `sbtc` is the dockerhub repostiory with all the devnev images, and the following `sbtc` is the tag for the actual sbtc image.

What might be better, is if we changed each image to have the name `blockstack/sbtc:devenv-[service]` so that it's clear this is for the devenv. Thoughts?

sbtc dockerhub repo: https://hub.docker.com/r/blockstack/sbtc

.github/workflows/publish-to-dockerhub.yml
```diff
@@ -0,0 +1,79 @@
...
+        with:
+          # Note, this is a little hacky because it requires each service that we 
+          # have to use the same folder structure.
+          context: "{{ defaultContext }}:/devenv/${{ matrix.service }}/docker"
+          file: ./Dockerfile
+          push: true
+          # TODO: Set up docker publishing to use the tag name when publishing and to do
+          # something else when triggered manually. $\{{ github.event.release.tag_name }}
+          tags: blockstack/sbtc:${{ matrix.service }}
```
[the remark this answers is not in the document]
AshtonStephens: So then these tags could be `blockstack/sbtc:devenv-[service]`?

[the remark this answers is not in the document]
CAGS295: Not the most experienced in tagging. It looks good to me.

.github/workflows/publish-to-dockerhub.yml
wileyj: same question as @CAGS295  - is there a reason `sbtc` can't be added to this matrix that i'm not seeing here?
the build step looks compatible.

.github/workflows/publish-to-dockerhub.yml
```diff
@@ -0,0 +1,79 @@
...
+          context: "{{ defaultContext }}"
+          file: ./devenv/sbtc/docker/Dockerfile
+          # TODO: Set up docker publishing to use the tag name when publishing and to do
+          # something else when triggered manually. $\{{ github.event.release.tag_name }}
+          tags: blockstack/sbtc:devenv-sbtc
+          push: true
+          platform: amd64
+
+  push_services_to_registry:
+    strategy:
```
wileyj, line 41:
```suggestion
    strategy:
      fail-fast: false
      max-parallel: 4 # or 8,10 etc
```

wileyj (approved):
some minor changes that might be good to make, but the workflow isn't complicated so I'll approve it. suggested changes/questions are not blockers

.github/workflows/publish-to-dockerhub.yml
CAGS295: `if: ${{ matrix.service }} != sbtc` No?

friedger:
Is this good to merge?

.github/workflows/publish-to-dockerhub.yml
CAGS295: I understand that the check means to only run if it is not sbtc. Otherwise, I don't understand the purpose of the if condition.

friedger:
@AshtonStephens This is still helpful for devenv when using nakamoto. Can we have docker images for the current state of stacks-blockchain node?

PR #288, opened 2023-10-15
Don't panic while polling for stacks blocks if the block isn't present.

AshtonStephens:
## Summary of Changes

This PR allows the stacks polling loop to handle 404 errors.

Currently we never poll the stacks blocks because romeo errors immediately when it encounters a 404 error.
This fixes that error.

### Risks

We could loop indefinitely on a request failure, but since this loop is the polling loop, that's the best place for there to be an infinite loop.

Tested locally against testnet with a running romeo binary.

```
./target/debug/romeo -c config.json
```
Where the config was 
```json
{
  "stacks_network": "testnet",
  "bitcoin_network": "testnet",
  "state_directory": "./state",
  "mnemonic": "******",
  "stacks_node_url": "https://api.testnet.hiro.so",
  "bitcoin_node_url": "http://blockstack:blockstacksystem@bitcoind.testnet.stacks.co:18332",
  "electrum_node_url": "ssl://blockstream.info:700",
  "contract_name": "asset",
  "hiro_api_key": "******"
}
```

We should test with real deposits and withdrawals on testnet.

## Checklist:

### DIFF
--- a/romeo/src/stacks_client.rs
+++ b/romeo/src/stacks_client.rs
@@ -266,18 +266,20 @@ impl StacksClient {
 		block_height: u32,
 	) -> anyhow::Result<Vec<StacksTransaction>> {
 		let res: Value = loop {
-			let inner_res: Value = self
+			let maybe_response: Result<Value, Error> = self
 				.send_request(|| {
 					self.http_client
 						.get(self.block_by_height_url(block_height))
 						.build()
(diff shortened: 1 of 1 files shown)


friedger (approved):
Looks good

stjepangolemac (approved):

Issue #261, opened 2023-10-11
Bug: mint on testnet fails due to btc block not yet processed


radicleart:
### Expected Behavior

Mint - in response to btc deposit - succeeds.

### Actual Behavior

Mint fails with error u6 (ERR-HEADER-HEIGHT-MISMATCH)

See [mint tx](https://explorer.hiro.so/txid/0x0b4d7fcebffe0dddd45825c77a5d428c2c81a63092e3d9a6f0d8ac0104840068?chain=testnet)

### Steps to Reproduce

1. run Romeo on testnet
2. send deposit using bridge or sbtc-cli

friedger:
Did you deploy the `-deploy` contract in testnet deployment plan?

radicleart:
Debug = false

See https://explorer.hiro.so/txid/0x53fa79edb8f86abc2e20421d289fa951deb624d9e11328d92f1cd437ef92df97?chain=testnet

radicleart:
What if we increase number of confirmations required on bitcoin before minting/burning currently its 1 ?

AshtonStephens:
@radicleart I'm going to implement this by allowing the mint to fail when there's not enough information and performing a retry later in the state machine.

Do you have a clear idea of how I might test this outside of Testnet?

stjepangolemac:
Some ideas:

- Either introduce new events for mint / burn failure or extend existing ones
- When one of them fails reset its state in the same way it was before the first try, that way it does it again on next block

Issue #101, opened 2023-08-31
sbtc-cli generates different stx address than clarinet


friedger:
Using the mnemonic defined in `romeo/asset-contracts/settings/Devnet.toml` for `sbtc generate-from mnemonic "...."` result in a different stx address

clarinet generates 
```
# stx_address: ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM
# btc_address: mqVnk6NPRdhntvfm4hh9vvjiRkFDUuSYsH
```
sbtc generates:
```
  "stacks_address": "ST24M0AD5N2QXBXC8HKFVZVYJQP4VFMA0VVYWN4WQ",
```

stjepangolemac:
This might have something to do with the difference in key derivation paths between Bitcoin and Stacks addresses. Maybe we need to look into stacks.js and see how they do it, also why they do it.

friedger:
[stacks.js](https://github.com/hirosystems/stacks.js/blob/b2a5f96fd24e8da7cb9b4e1cf4d7b654f6e5b00c/packages/wallet-sdk/src/derive.ts#L12) uses:
``` 
const DATA_DERIVATION_PATH = `m/888'/0'`;
const WALLET_CONFIG_PATH = `m/44/5757'/0'/1`;
const STX_DERIVATION_PATH = `m/44'/5757'/0'/0`;
```

friedger:
stbc-cli does not use any derivation path

stjepangolemac:
I'll need to figure out how to use derivation paths with bdk. Do you know anything about that?

stjepangolemac:
Fixed in https://github.com/stacks-network/sbtc/issues/119

Issue #100, opened 2023-08-31
[sBTC DR] Make Romeo more resilient to recoverable errors


stjepangolemac:
We have a lot of unwraps in the codebase that might be recoverable, or handled in a clearer way. Also, we've detected a crash caused by a closed HTTP connection to the Esplora service. That can be retried.

Issue #11, opened 2023-07-31
Define end-2-end testing scenario for Mini


netrome:
From [testing dialogue](https://github.com/stacks-network/stacks-blockchain/discussions/3787#discussioncomment-6577287)

To prepare for e2e integration testing for Mini, we'd like to define a first scenario.  For example,

1. configure and run a bitcoin and a stacks node
2. configure and run signers on different machines
3. wait for this...
4. observe this...
5. shutdown the machines

Once this is defined we can collaborate with the testing group to set it up.

### Definition of done
An initial testing scenario for Mini is set up

CAGS295:
Is [this](https://github.com/Trust-Machines/stacks-sbtc/tree/main/sbtc-mini) mini?

netrome:
> Is [this](https://github.com/Trust-Machines/stacks-sbtc/tree/main/sbtc-mini) mini?

Yes. We have docs in progress. It's being rebranded as sBTC 0.1/Developer release.

CAGS295:
**WIP**

1. Select a network target and network settings.
2. Start a global orchestrator using DevnetJS.
3. Deploy smart contract tests (if ephemeral network).
4. Await or walkthrough until preconditions are met. e.g. epochs?
5. Execute tests.
    - Non-consensus tests run in parallel or in lockstep.
    - Consensus tests run sequentially.
7. Destroy smart contract tests (if feasible).
8. Terminate the orchestrator.

CAGS295:
@netrome can you point me to the current e2e/integration solution for sBTC?

netrome:
> @netrome can you point me to the current e2e/integration solution for sBTC?

There isn't one. This is the first stab at it.

bitfari:
Multiple issues missing, for example: we need to store the addresses of generated nodes, plus have configurable variables to allot different waiting times across envs (test, dev, main). Currently drafting something a bit more concrete.

CAGS295:
> Multiple issues missing, for example: we need to store the addresses of generated nodes, plus have configurable variables to allot different waiting times across envs (test, dev, main). Currently drafting something a bit more concrete.

It sounds like I've got the scope of the tests wrong. Your input makes me think that you actually need to deploy a network on demand and make sure through tests that consensus works as expected. Put simply, do the integration tests involve a single node or many nodes?

CAGS295:
[...]

PR #290, opened 2023-10-16
Add wait period between receiving deposit request and minting.

AshtonStephens:
## Summary of Changes

This PR aims to resolve https://github.com/stacks-network/sbtc/issues/261, in which we were attempting to mint deposit transactions that hadn't yet been mined by the chain.

This PR resolves this bug by adding a `Scheduled` state to each deposit request in which the deposit requests are left alone until the next stacks block.

### Risks

If something went wrong this could result in deposits never getting fulfilled, but that seems unlikely given the simplicity of this logic.

Not tested, though it should be. This is a fairly simple change and will never be on mainnet.

At the moment it's riskier not to include this PR or spend additional resources testing it than it is to accept as is.

Should be tested as part of a testnet program.

## Checklist:

### DIFF
--- a/romeo/src/state.rs
+++ b/romeo/src/state.rs
@@ -22,7 +22,7 @@ use crate::{
 
 /// The delay in blocks between receiving a deposit request and creating
 /// the deposit transaction.
-const MINT_TRANSACTION_DELAY_BLOCKS: u32 = 1;
+const STX_TRANSACTION_DELAY_BLOCKS: u32 = 1;
 
 /// Romeo internal state
 #[derive(Debug, serde::Serialize, serde::Deserialize)]
@@ -485,7 +485,7 @@ impl State {
 							// one we make ourselves resilient to mining delays
 							// without complex logic.
 							let scheduled_block_height = *stacks_block_height
-								+ MINT_TRANSACTION_DELAY_BLOCKS;
+								+ STX_TRANSACTION_DELAY_BLOCKS;
 
 							deposit.mint =
 								Some(TransactionRequest::Scheduled {
@@ -517,7 +517,7 @@ impl State {
 							None => {
 								let scheduled_block_height =
 									*stacks_block_height
-										+ MINT_TRANSACTION_DELAY_BLOCKS;
+										+ STX_TRANSACTION_DELAY_BLOCKS;
 
 								withdrawal.burn =
 									Some(TransactionRequest::Scheduled {
--- a/romeo/src/state.rs
+++ b/romeo/src/state.rs
@@ -479,9 +479,11 @@ impl State {
 				let deposit_tasks = deposits.iter_mut().filter_map(|deposit| {
 					match deposit.mint.as_mut() {
 						None => {
-							// We often receive the deposit before the transaction is actually mined.
-							// By scheduling the transaction for a block later than the current one we
-							// make ourselves resilient to mining delays without complex logic.
+							// We often receive the deposit before the
+							// transaction is actually mined. By scheduling the
+							// transaction for a block later than the current
+							// one we make ourselves resilient to mining delays
+							// without complex logic.
 							let scheduled_block_height = *stacks_block_height
 								+ MINT_TRANSACTION_DELAY_BLOCKS;
 							deposit.mint =
@@ -496,8 +498,9 @@ impl State {
 							block_height,
 						}) => {
 							if stacks_block_height.ge(&block_height) {
-								// Only initiate the mint task if the current stacks block is
-								// or is after the stacks block for which the mint is scheduled.
+								// Only initiate the mint task if the current
+								// stacks block is or is after the stacks block
+								// for which the mint is scheduled.
 								deposit.mint =
 									Some(TransactionRequest::Created);
 								info!(
--- a/romeo/src/state.rs
+++ b/romeo/src/state.rs
@@ -22,7 +22,7 @@ use crate::{
 
 /// The delay in blocks between receiving a deposit request and creating
 /// the deposit transaction.
-const MINT_TRANSACTION_DELAY_BLOCKS: u32 = 3;
+const MINT_TRANSACTION_DELAY_BLOCKS: u32 = 1;
 
 /// Romeo internal state
 #[derive(Debug, serde::Serialize, serde::Deserialize)]
@@ -486,17 +486,20 @@ impl State {
 							// without complex logic.
 							let scheduled_block_height = *stacks_block_height
 								+ MINT_TRANSACTION_DELAY_BLOCKS;
+
 							deposit.mint =
 								Some(TransactionRequest::Scheduled {
 									block_height: scheduled_block_height,
 								});
+
 							debug!("Scheduled deposit {} for minting on stacks block height {}.",
 								deposit.info.txid, scheduled_block_height);
+
 							None
 						}
 						Some(TransactionRequest::Scheduled {
 							block_height,
-						}) if (block_height <= stacks_block_height) => {
+						}) if (*block_height <= *stacks_block_height) => {
 							// Only initiate the mint task if the current
 							// stacks block is or is after the stacks block
 							// for which the mint is scheduled.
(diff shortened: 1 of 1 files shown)


romeo/src/state.rs
```diff
@@ -469,13 +473,44 @@ impl State {
...
+									block_height: scheduled_block_height,
+								});
+							info!("Scheduled deposit {} for minting on stacks block height {}.",
+								deposit.info.txid, scheduled_block_height);
+							None
+						}
+						Some(TransactionRequest::Scheduled {
+							block_height,
+						}) => {
+							if stacks_block_height.ge(&block_height) {
```
CAGS295, line 500: What happens if the stack blocks are full and the mint is not included at the expected height?

romeo/src/state.rs
```diff
@@ -469,13 +473,44 @@ impl State {
...
+								deposit.mint =
+									Some(TransactionRequest::Created);
+								info!(
+									"Created mint for {}.",
+									deposit.info.txid
+								);
+								Some(Task::CreateMint(deposit.info.clone()))
+							} else {
+								None
+							}
```
CAGS295, line 513: Consider flattening the matching pattern.

```suggestion
}) if stacks_block_height.ge(&block_height) => {
// Only initiate the mint task if the current
// stacks block is or is after the stacks block
// for which the mint is scheduled.
    deposit.mint = Some(TransactionRequest::Created);info!("Created mint for {}.", deposit.info.txid);
    Some(Task::CreateMint(deposit.info.clone()))
```

romeo/src/state.rs
```diff
@@ -469,13 +473,44 @@ impl State {
 			State::Initialized {
 				deposits,
 				withdrawals,
+				stacks_block_height,
 				..
 			} => {
```
CAGS295, line 478:
```suggestion
			} => {
    /// The delay in blocks between receiving a deposit request and
    /// creating the deposit transaction.
    const MINT_TRANSACTION_DELAY_BLOCKS: u32 = 1;

```

romeo/src/state.rs
```diff
@@ -20,6 +20,10 @@ use crate::{
 	task::Task,
 };
 
+/// The delay in blocks between receiving a deposit request and creating
+/// the deposit transaction.
+const MINT_TRANSACTION_DELAY_BLOCKS: u32 = 1;
+
```
CAGS295, line 26: Reduce the constant's scope.
```suggestion

```

CAGS295 (dismissed):

romeo/src/state.rs
AshtonStephens: As much as I love this, clippy and fmt will hate it.

AshtonStephens: In fact, this what I originally had, except you accidentally deleted the `None` part.

romeo/src/state.rs
AshtonStephens: Then we're SOL, this is not an ideal solution but it's one that'll fix the issue 99/100 times.

AshtonStephens: There's a world in which we hit the transaction with the clarity read-only function `verify-txid-exists-on-burn-chain` and wait for each block until that verification passes - that's the *right* way to do this.

We said we'd have on testnet a while ago so it's better to make a fast decent fix that will likely solve the problem than introduce complexity. Especially since this code isn't going to be built upon for the actual sBTC release.

If we have this problem again we can refer to these comments.

romeo/src/state.rs
AshtonStephens: Following the constant standards in the other files,  I think it makes more sense to leave this where it is:

[`BLOCK_POLLING_INTERVAL` in the `stacks_client`](https://github.com/stacks-network/sbtc/blob/ebdcc7bc44aac6628bff8c9b8b5924f0589c0be7/romeo/src/stacks_client.rs#L34)

romeo/src/state.rs
AshtonStephens: Responded below.

romeo/src/state.rs
```diff
@@ -469,13 +473,44 @@ impl State {
...
+						Some(TransactionRequest::Scheduled {
+							block_height,
+						}) => {
+							if stacks_block_height.ge(&block_height) {
+								// Only initiate the mint task if the current
+								// stacks block is or is after the stacks block
+								// for which the mint is scheduled.
+								deposit.mint =
+									Some(TransactionRequest::Created);
+								info!(
```
stjepangolemac, line 506: I'd convert this to a `trace!` as it's not too important to see in normal operation, or `debug!`.

stjepangolemac (dismissed):
This is a very nice implementation that we've mentioned before but thought it'll be too complex without being perfect.

I stand corrected as this is nice, short, and should cover 99% of cases. The only thing we need to think about is what to set `MINT_TRANSACTION_DELAY_BLOCKS` to. I think @friedger will have a nice number in his mind. I suggest 5?

romeo/src/state.rs
CAGS295: Really? https://github.com/stacks-network/sbtc/actions/runs/6538570444/job/17754800452?pr=293

AshtonStephens:
Okay, I'm going to make some of these changes. @CAGS295 proved my linter claims wrong!

A followup commit coming shortly..

romeo/src/state.rs
CAGS295: SOL?

Fine with me. I wanted an answer for the record in case I missed something.

romeo/src/state.rs
AshtonStephens: Get rekt! And, the part I didn't realize here was the additional pattern match here is what you were talking about:

```rust
Some(TransactionRequest::Scheduled {
	block_height, 
}) if (block_height <= stacks_block_height) => { // <------ Part of the pattern matching! Cool!
...
}
```

Thanks Carlos, this is much better, 10/10.

romeo/src/state.rs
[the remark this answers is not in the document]
AshtonStephens: Done.

romeo/src/state.rs
AshtonStephens: Done.

CAGS295 (approved):

romeo/src/state.rs
```diff
@@ -12,14 +12,18 @@ use sbtc_core::operations::{
...
 
 use crate::{
 	config::Config,
 	event::{Event, TransactionStatus},
 	task::Task,
 };
 
+/// The delay in blocks between receiving a deposit request and creating
+/// the deposit transaction.
+const MINT_TRANSACTION_DELAY_BLOCKS: u32 = 3;
```
friedger, line 25: It should be 1. The experience would be bad if we wait always 5 blocks for no reasons.

friedger: We just need to wait until the stacks blockchain sees the bitcoin block. That will be in the next stacks block the latest.

romeo/src/state.rs
friedger: We only want to prevent broadcasting too early. We have an extra event for inclusion once the tx was confirmed.

I don't see a problem here.

romeo/src/state.rs
friedger, line 25:
```suggestion
const MINT_TRANSACTION_DELAY_BLOCKS: u32 = 1;
```

stjepangolemac: If this is set to 1, would it make sure we can handle 5 flash blocks in a row? I'm not sure how common that is.

friedger: The height is the stacks height, therefore, we can handle as many flash blocks as there are.
Btc block #100 and Stx block #10 are published.
Then Btc block #101 is publish with deposit request, we wait for STX block #11 before broadcasting minting because btc block #101 is not visible while building block #11. 

(Maybe my numbers are off by 1 but we don't have to worry about the number of flash blocks. When we wait for block #11 all the btc block #101, 102, 103,... would be covered in the same way)

stjepangolemac:
Set the constant back to `1` and delayed withdrawals too.

friedger (approved):
LGTM

stjepangolemac (approved):

PR #152, opened 2023-09-20
[sBTC DR] Introduce property-based tests for asset.clar

moodmosaic:
# Introduce property-based tests for asset.clar

Invariant properties allow you to verify that certain conditions always hold true—not only that nothing bad ever happens, but also that something good is always possible. 

This technique became recently popular in the EVM community under the term 'property-based fuzzing' and (to my knowledge) was pioneered by dapp.tools and echidna, and became mainstream by foundry.<sup>1</sup>

This PR contains an encoding of this technique using [fast-check](https://github.com/dubzzz/fast-check) and clarinet, to falsify user-defined invariants (properties) instead of looking just for crashes like a traditional fuzzer. 

```
wallet_3         transfer wallet_7   1140351122 (discarded)
wallet_2 get-total-supply                     0
deployer             mint   faucet   1119048227 bitcoin-tx 0168ee41db8a4766efe02bba1ebc0de320bc1b0abb7304f5f104818a9dd721cf
  faucet get-total-supply            1119048227
deployer             mint wallet_7   2133632941 bitcoin-tx 255dcfd00b04456288b5aacc5275835add15c89067f082952331b7fc1b87a63c
deployer             mint wallet_6   1910196597 bitcoin-tx 255dcfd00b04456288b5aacc5275835add15c89067f082952331b7fc1b87a63c
deployer             mint wallet_4    768982343 bitcoin-tx 255dcfd00b04456288b5aacc5275835add15c89067f082952331b7fc1b87a63c
  faucet         transfer wallet_4   1299487135 (expected, non-owner)
wallet_3      get-balance wallet_6   1910196597
wallet_4         transfer wallet_2   1114063953 (discarded)
wallet_4 get-total-supply            5931860108
wallet_3         transfer wallet_2   1497519370 (discarded)
wallet_4         transfer deployer    169030480
wallet_8 get-total-supply            5931860108
wallet_1 get-total-supply            5931860108
wallet_5 get-total-supply            5931860108
wallet_6 get-total-supply            5931860108
deployer             burn   faucet   1058352425 bitcoin-tx 0168ee41db8a4766efe02bba1ebc0de320bc1b0abb7304f5f104818a9dd721cf
wallet_3         transfer wallet_3    806137080 (discarded)
deployer             mint wallet_8    407684029 bitcoin-tx 255dcfd00b04456288b5aacc5275835add15c89067f082952331b7fc1b87a63c
[...]

PR #352, opened 2023-11-05
[sBTC DR] property-based tests for asset.clar

moodmosaic:
Continuation of #152. This pull request covers `get/set-bitcoin-wallet-public-key`, and keeps the console output aligned.

```
  ST2C...K9AG              get-total-supply                      128
  ST1P...GZGM set-bitcoin-wallet-public-key 0xf6...8e3e
  ST1P...GZGM                   get-balance ST2R...P2VB            0
  ST2R...P2VB              get-total-supply                      128
! ST2N...87ND set-bitcoin-wallet-public-key 0xa6...1fca                                      (expected, non-owner)
  ST1P...GZGM set-bitcoin-wallet-public-key 0x52...fb67
! ST2R...P2VB set-bitcoin-wallet-public-key 0x24...555b                                      (expected, non-owner)
! ST1S...YPD5 set-bitcoin-wallet-public-key 0x2b...eed0                                      (expected, non-owner)
! ST1S...YPD5 set-bitcoin-wallet-public-key 0xb2...a7fb                                      (expected, non-owner)
  ST1P...GZGM set-bitcoin-wallet-public-key 0x71...d2d4
! ST3A...GCS0 set-bitcoin-wallet-public-key 0x59...cbdc                                      (expected, non-owner)
! ST2J...5NNC set-bitcoin-wallet-public-key 0x0a...cc99                                      (expected, non-owner)
  STNH...MAZ6              get-total-supply                      128
  STNH...MAZ6              get-total-supply                      128
  ST1S...YPD5              get-total-supply                      128
! ST3N...1XCP                      transfer STNH...MAZ6    299493701                         (discarded, not enough funds)
! ST2J...5NNC                      transfer ST2J...5NNC   1390308638                         (discarded, sender can not be the receiver)
! ST2J...5NNC                      transfer ST2J...5NNC   1390308638                         (discarded, not enough funds)
! ST1S...YPD5                      transfer ST1P...GZGM   1655155279                         (discarded, not enough funds)
! ST3P...YGKJ                      transfer ST3P...YGKJ   1886821251                         (discarded, sender can not be the receiver)
! ST3P...YGKJ                      transfer ST3P...YGKJ   1886821251                         (discarded, not enough funds)
! ST1P...GZGM                          mint ST2J...5NNC           71 bitcoin-tx  255d...a63c (expected, same bitcoin-tx)
! ST3P...YGKJ                      transfer ST2R...P2VB   2090993235                         (discarded, not enough funds)
! ST1P...GZGM                      transfer ST2J...5NNC    292423422                         (discarded, not enough funds)
```

## Summary of Changes

Adds property-based tests covering `get/set-bitcoin-wallet-public-key`.

### Risks

No risks associated with this PR. Existing tests aren't modified, only new tests added ([append-only tests](https://blog.ploeh.dk/2013/04/02/why-trust-tests/#b23efe23c1574f2b8ff85fb3c529ec9d)).

Once you `cd` into `romeo/asset-contract` directory, run `npm install`, then you can run the tests via

```bash
npm test
```

## Checklist:

(Let me know if there's a style guideline of this project that I can follow, and any changes to the documentation that I should make. There are no dependent changes in downstream modules.)

### DIFF
--- a/romeo/asset-contract/tests/asset_SetBitcoinWalletPublicKeyCommand.ts
+++ b/romeo/asset-contract/tests/asset_SetBitcoinWalletPublicKeyCommand.ts
@@ -1,0 +1,59 @@
+import {
+  AssetCommand,
+  Real,
+  Stub,
+  shortenString
+} from "./asset_CommandModel.ts";
+
+import { tx } from "@hirosystems/clarinet-sdk";
+import { Cl } from "@stacks/transactions";
+
+import { expect } from "vitest";
+
+export class SetBitcoinWalletPublicKeyCommand implements AssetCommand {
+  readonly sender: string;
+  readonly pubKey: Uint8Array;
+
+  constructor(
+    sender: string,
+    pubKey: Uint8Array,
+  ) {
+    this.sender = sender;
+    this.pubKey = pubKey;
+  }
+
+  check(_: Readonly<Stub>): boolean {
+    // Can run if sender is the deployer. This is filtered in the generator.
+    return true;
+  }
+
+  run(model: Stub, real: Real): void {
+    const block = real.simnet.mineBlock([
+      tx.callPublicFn(
+        "asset",
+        "set-bitcoin-wallet-public-key",
+        [
+          Cl.buffer(this.pubKey),
+        ],
+        this.sender,
+      ),
+    ]);
+
+    expect(block[0].result).toBeOk(Cl.bool(true));
+
+    console.log(
+      `✓ ${shortenString(this.sender).padStart(8, " ")} ${"set-bitcoin-wallet-public-key".padStart(29, " ") } ${shortenString(uint8ArrayToHexString(this.pubKey))}`
+    );
+  }
+
+  toString() {
+    // fast-check will call toString() in case of errors, e.g. property failed.
+    // It will then make a minimal counterexample, a process called 'shrinking'
+    // https://github.com/dubzzz/fast-check/issues/2864#issuecomment-1098002642
+    return `${this.sender} set-bitcoin-wallet-public-key ${uint8ArrayToHexString(this.pubKey)}`;
+  }
+}
+
+function uint8ArrayToHexString(uint8Array: Uint8Array): string {
+  return '0x' + Array.from(uint8Array).map(byte => byte.toString(16).padStart(2, '0')).join('');
+}
--- a/romeo/asset-contract/tests/asset_SetBitcoinWalletPublicKeyCommand.ts
+++ b/romeo/asset-contract/tests/asset_SetBitcoinWalletPublicKeyCommand.ts
@@ -41,6 +41,8 @@ export class SetBitcoinWalletPublicKeyCommand implements AssetCommand {
 
     expect(block[0].result).toBeOk(Cl.bool(true));
 
+    model.bitcoinWalletPublicKey = this.pubKey;
+
     console.log(
       `✓ ${shortenString(this.sender).padStart(8, " ")} ${"set-bitcoin-wallet-public-key".padStart(29, " ") } ${shortenString(uint8ArrayToHexString(this.pubKey))}`
     );
--- a/romeo/asset-contract/tests/asset_SetBitcoinWalletPublicKeyCommand_NonOwner.ts
+++ b/romeo/asset-contract/tests/asset_SetBitcoinWalletPublicKeyCommand_NonOwner.ts
@@ -1,0 +1,59 @@
+import {
+  AssetCommand,
+  Real,
+  Stub,
+  shortenString
+} from "./asset_CommandModel.ts";
+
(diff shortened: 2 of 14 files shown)


romeo/asset-contract/tests/asset_SetBitcoinWalletPublicKeyCommand.ts
```diff
@@ -0,0 +1,61 @@
...
+    sender: string,
+    pubKey: Uint8Array,
+  ) {
+    this.sender = sender;
+    this.pubKey = pubKey;
+  }
+
+  check(_: Readonly<Stub>): boolean {
+    // Can run if sender is the deployer. This is filtered in the generator.
+    return true;
```
friedger, line 27: Instead of comment, it might make sense to enforce the logic in the generator here as well.

romeo/asset-contract/tests/asset_SetBitcoinWalletPublicKeyCommand_NonOwner.ts
```diff
@@ -0,0 +1,59 @@
...
+    sender: string,
+    pubKey: Uint8Array,
+  ) {
+    this.sender = sender;
+    this.pubKey = pubKey;
+  }
+
+  check(_: Readonly<Stub>): boolean {
+    // Can run if sender is not the deployer. This is filtered in the generator.
+    return true;
```
friedger, line 27: Logic of generator should be enforced here for better understanding.

friedger (commented):
Looks good. Does it make sense to enforce the logic of the generator in the commands? It makes the commands easier to read/understand.

friedger (approved):

romeo/asset-contract/tests/asset_SetBitcoinWalletPublicKeyCommand.ts
moodmosaic: We should *probably* do it... However, there is a catch when doing that:

Say that the command runs 100 times.

If you enforce the logic there in the `check` call, you'll be [discarding](https://www.cse.chalmers.se/~rjmh/QuickCheck/manual_body.html#6) the test-run if the condition is not met. For example, if the sender is not the `deployer` 30% of the times, you'll end up running the command 30% of the times less.

If you filter in the generator (as-is right now), you guarantee to always run the command with a `deployer` as the sender. I'll double-check if my theory is correct with fast-check though, and will update those `checks` if that's not the case.

---

In general, I'm planning to move us away from those boilerplate-ish commands, and define them on the fly, right next to where the generators are defined. [Technically it seems possible](https://github.com/dubzzz/fast-check/discussions/3026) so all that's left is to try it.

CAGS295: I agree with moodmosaic. Furthermore, in my mind, `check` is good for 'managing' model transitions while the generators model input domain.

romeo/asset-contract/tests/asset_SetBitcoinWalletPublicKeyCommand_NonOwner.ts
CAGS295: I disagree. This would skew command selection, as explained by Moodmosaic. Ambiguity in input generation should be resolved 'at' the generators.

CAGS295 (approved):

CAGS295:
Question. Are we running a fast-check in CI yet?

romeo/asset-contract/tests/asset_SetBitcoinWalletPublicKeyCommand.ts
moodmosaic: In fast-check, `check` is the place where you decide whether the command (with its generated data) should run or not, given the current state of the model.

---

From the [docs](https://fast-check.dev/docs/advanced/model-based-testing/#define-the-commands):


| Syntax           | Description                                                             |
| ---------------- | ----------------------------------------------------------------------- |
| check(model)     | Ensure that the model is in the appropriate state to execute the action |
| run(model, real) | Execute the action                                                      |
| toString()       | Serialize the command for error reports                                 |

romeo/asset-contract/tests/asset_SetBitcoinWalletPublicKeyCommand_NonOwner.ts
moodmosaic: See https://github.com/stacks-network/sbtc/pull/352#discussion_r1385614571. But in reality, what @friedger writes makes sense to me; it'd be much easier if everything is in one place.

Perhaps when we switch to [defining commands on-the-fly](https://github.com/dubzzz/fast-check/discussions/3026) things will be more clear, since both the generator and the command will be in the 'same' place.

moodmosaic:
> Question. Are we running fast-check in CI yet?

fast-check tests run as part of `npm test` now.

